PR #186, opened 2014-09-25
Support for accepting paste submissions after deadline

jamox:
When paste is submitted after deadline and all test have passed, no points are given, and following notification is shown:
screen shot 2014-09-25 at 23 46 16


jamox:
Nope

PR #223, opened 2015-01-25
All e-mails: put sender in Reply-To instead of From.

mpartel:
Hopefully this makes recipient servers less likely to flag our e-mails as spam.


jamox:
LGTM

Issue #191, opened 2014-10-11
Hide review feature if not explicitly enabled


mpartel:
Web interface and IDE should hide everything to do with reviews if `course_options.yml` doesn't enable the feature explicitly. This prevents users from requesting reviews when nobody is actually committed to handling them.

PR #228, opened 2015-02-07
Add option to advise clients to disable code review requests.

jamox:


josalmi:
Connects to #191

jamox:
Should this default to false as described in #191?

mpartel:
I think so.

Note that the migration should first default it to true to set it to true on all existing exercises and then change the default to false.

LGTM otherwise.

jamox:
Updated migration. Set default value to false and update all existing records to have value true.

mpartel:
Feel free to merge after addressing the above two comments.

PR #231, opened 2015-02-11
Better formatting for test results

josalmi:
http://tmc.mooc.fi/mooc/paste/CJGmLgMXJ4Jdnf-zHvrYow#testResults


jamox:
We should add some formatting to make the pre stand out less. maybe remove border and inherit background color.

jamox:
Used this as a base for PR #233

PR #233, opened 2015-02-16
Error messages in pre so that viewing indentation etc would be easier

jamo:


mpartel:
Have you tested that very long messages don't overflow in some nasty way?

jamox:
Yeah, the formatting is not ideal with this either.
But anyways it is not nice either...

PR #239, opened 2015-02-22
Added option to disable run tests locally action

jamox:


jamox:
(All tests has passed, though no tests were added/modified...)

PR #241, opened 2015-02-26
Only give points that are available from the exercise

josalmi:
We don't want this to happen: http://u.tktl.fi/png/e2329b31ce2ee722bc2196b6e1832669


mpartel:
There will always be many ways to cheat but easy fixes like this seem reasonable anyway. Thanks!

PR #242, opened 2015-02-28
Revert "Only give points that are available from the exercise"

jamox:
Reverts testmycode/tmc-server#241


mpartel:
What went wrong?

jamox:
Oh, right, no points were given at all.

mpartel:
And tests passed??

jamox:
It appears that nobody ever run tests for the PR...

PR #244, opened 2015-03-06
WIP: Only give points that are available from the exercise

josalmi:
We don't want this to happen: http://u.tktl.fi/png/e2329b31ce2ee722bc2196b6e1832669

### DIFF
--- a/lib/test_run_grader.rb
+++ b/lib/test_run_grader.rb
@@ -86,10 +86,11 @@ module TestRunGrader
     user = submission.user
     exercise = submission.exercise
     course = exercise.course
+    available_points = exercise.available_points.to_a.map(&:name)
     awarded_points = AwardedPoint.course_user_points(course, user).map(&:name)
 
     points = []
-    for point_name in points_from_test_results(results) - review_points
+    for point_name in (available_points & points_from_test_results(results)) - review_points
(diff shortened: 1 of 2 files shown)


spec/lib/test_run_grader_spec.rb
```diff
@@ -81,6 +81,27 @@ def successful_results
     expect(points).not_to include('1.2')
   end
 
+  it 'should only award points which are available from the exercise' do
+    otherExercise = FactoryGirl.create(:exercise, course_id: @submission.course_id)
+    FactoryGirl.create(:available_point, exercise_id: otherExercise.id, name: '1.3')
+
+    results = [
+      {
+        'className' => 'MyTest',
```
kasper: Old hash-syntax.

mpartel:
I take it you ran all the tests this time? :)

lib/test_run_grader.rb
```diff
@@ -86,10 +86,11 @@ def self.award_points(submission, results, review_points)
     user = submission.user
     exercise = submission.exercise
     course = exercise.course
+    available_points = exercise.available_points.to_a.map(&:name)
```
jamox: Why `.to_a`?

jamox:
LGTM, except for the old hash syntax, can we change it? (or do we need to stick with it for the string keys)

PR #248, opened 2015-04-09
WIP: Generate course certificate after completion - V2

josalmi:


app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    return respond_access_denied('Authentication required') if current_user.guest?
+  end
+
+  def show
+    certificate = Certificate.find params[:id]
+    authorize! :read, certificate
+
+    visible_exercises = certificate.course.exercises.select { |e| e.points_visible_to?(certificate.user) }
+    total_available = AvailablePoint.course_points_of_exercises(certificate.course, visible_exercises).length
```
jamox: maybe `.count` instead of `.length`?

app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+  end
+
+  private
+  def certificate_params
+    params.require(:certificate).permit(:name, :user_id, :course_id)
+  end
+
+  def courses
+      @courses ||= Course
```
jamox: weird indentation

app/models/course.rb
```diff
@@ -236,6 +251,14 @@ def submissions_to_review
     submissions.where('(requests_review OR requires_review) AND NOT reviewed AND NOT newer_submission_reviewed AND NOT review_dismissed')
   end
 
+  def certificate_downloadable_for?(user)
+    user.administrator? ||
+      (!user.guest? &&
+       certificate_downloadable &&
+       (certificate_unlock_spec.nil? ||
+        UnlockSpec.new(exercises.first, ActiveSupport::JSON.decode(certificate_unlock_spec)).permits_unlock_for?(user)))
```
josalmi: Is there a better way? exercises.first is not nice!

app/views/certificates/new.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>New certificate</h1>
+<%= render partial: 'form', locals: {certificate: @certificate} %>
```
jamox: Is this used?

jamox:
Pretty happy with this. Two concerns though.

We don't store the cert we have created?
I think for the sake of really validating it we should have copies of each cert ever produced.

The cert is always regenerated.
When user creates a cert I would expect the date/time to be exactly the one he previously had, not changing for each request.
-- Say after two years student would like to get the cert that he has completed the course (two years ago) but now the cert states that he has completed the course now.


We allow user to have one cert for each course?
With current behaviour this isn't really a problem, but if above changes are done then this might be an issue for students getting cert once it is possible and later willing to get cert with better score

jamox:
Should we have a validate/url/key in the cert for people to validate the cert?
maybe a nice qr code?

PR #251, opened 2015-04-18
Fix 'undefined local variable' errors

MawKKe:
52c15d4 Changed some template bindings (get_binding), causing
rake background_daemon:init:install to fail with:

        'undefined local variable or method `name'...'
        'undefined local variable or method `env'...'

And so on. I renamed the template variables to reflect the changes in
the previous commit


mpartel:
Thanks!

I wonder if the underscore is because RuboCop thought the arguments were unused :/

PR #257, opened 2015-05-29
Add travis support

jruo:
Build log: https://travis-ci.org/TMC-teacher-tools/tmc-server/builds/64532459

Currently it runs tests from:
- spec/controllers
- spec/models
- spec/helpers
- spec/mailers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,23 @@
+sudo: true
+
+language: ruby
+rvm:
+  - 2.2.0
+
+git:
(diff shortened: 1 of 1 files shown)


josalmi:
LGTM :+1: GJ

jamox:
Thanks for your work.
I really like that we can run part of the tests in travis, though I'm a bit worried that this gives us a fake sign of security -- we run only the most trivial tests.

PR #263, opened 2015-06-02
Migrate courses to organizations

Termanty:
Because courses belong to organizations now, all course paths were reconfigured to include organization id (slug).


jamox:
Nice work!

The usermanual test is still failing but it is kinda acceptable as the changes for the workflow are not yet fully implemented. We'll need to be modifying this: https://github.com/testmycode/tmc-server/blob/master/spec/usermanual/instructors.html.erb once the organisations workflow is done.

PR #265, opened 2015-06-05
Teacher can disable and enable courses

jsopakar:


app/controllers/courses_controller.rb
```diff
@@ -116,4 +126,12 @@ def assign_show_view_vars
   def set_organization
     @organization = Organization.find_by(slug: params[:organization_id])
   end
+
+  def disable_course(disabled)
+    @course = Course.find(params[:id])
+    @course.disabled = disabled
+    notice_action = disabled ? 'disabled' : 'enabled'
+    notice_message = (@course.save ? 'Course was successfully' : 'Course could not be') + " #{notice_action}."
+    redirect_to(organization_course_path(@organization, @course), notice: notice_message)
```
jamox: Is there a need to have the the course as an instance variable, isn't a local variable good enough

jruo: Fixed

app/controllers/courses_controller.rb
```diff
@@ -89,6 +89,16 @@ def create
...
   end
 
+  def enable
+    authorize! :teach, @organization
+    disable_course(false)
+  end
+
+  def disable
+    authorize! :teach, @organization
+    disable_course(true)
```
jamox: The `disable_course(false)` seems kinda unclear, maybe
`enable_course` and `disable_course` or just
`set_disabled_status(:disabled)` and `set_disabled_status(:enabled)`

Even thought the disable field is clearly boolean value, use of the rails 4 enum might make it syntactically nicer
we could just have 
`.disable!`, `.enable!` and `.disabled?` and `.enabled?`

PR #266, opened 2015-06-05
Teachers can list courses as json, organization front page shows courses

jruo:


config/routes.rb
```diff
@@ -17,6 +17,7 @@
       member do
         get 'refresh'
         post 'refresh'
+        get 'courses.json', to: 'courses#show_json', format: 'json', as: 'one_course_json'
```
jamox: is the `.json` required? (we have also the format)

so this would map to
`/courses/:id/course.json`? why not just `/courses/:id.json`

Termanty: Good catch. That `.json` is not needed.

That mapping to `/courses/:id/course.json` is because netbeans tmc-plugin adds `/courses.json` to request. Now student gets to select only one course.

Your earlier suggestion (magic url) is then part of little bit bigger changes where enrolment is coming to play.

app/controllers/courses_controller.rb
```diff
@@ -57,6 +57,19 @@ def show
     end
   end
 
+  # Method for teacher to give a single course for students to select.
+  def show_json
+    course = [Course.find(params[:id])]
+    authorize! :read, course
+    return respond_access_denied('Authentication required') if current_user.guest?
```
jamox: Not a huge fan of `if current_user.quest?`

jamox:
Usermanual needs to be updated, but after this effort is done (one test of it is failing)

PR #267, opened 2015-06-05
Gui deadline support

FeisEater:


jamox:
There is still some work for the UI of deadlines (e.g. date selector for the deadline, exercise specific deadline options).

Merging this now and these will be added in future PR.

PR #268, opened 2015-06-08
Course create from template visibility help

Termanty:


app/views/course_templates/index.html.erb
```diff
@@ -20,13 +21,21 @@
         <td><%= course_template.description %></td>
         <td><%= course_template.material_url %></td>
         <td><%= course_template.source_url %></td>
+        <td><%= course_template.expires_at %></td>
         <td><%= link_to 'Edit', edit_course_template_path(course_template) %></td>
         <td><%= link_to 'Destroy', course_template, method: :delete, data: { confirm: 'Are you sure?' } %></td>
+        <td><% if course_template.hidden %>
```
jamox: Maybe move the if to its own line

jsopakar: Fixed

config/routes.rb
```diff
@@ -1,10 +1,12 @@
 TmcServer::Application.routes.draw do
+  get 'org/:organization_id/course_templates', to: 'course_templates#list_for_teachers', as: 'list_course_templates'
+  get 'org/:organization_id/course_templates/:id', to: 'course_templates#prepare_course', as: 'prepare_course'
```
jamox: not a big fan of manual matchers; we probably could turn these into members of the ` resources :organizations, path: 'org' do`

app/controllers/courses_controller.rb
```diff
@@ -82,13 +75,19 @@ def create
...
+        refresh_course(@course)
+        format.html { redirect_to(organization_course_help_path(@organization, @course), notice: 'Course was successfully created.') }
       else
         format.html { render action: 'new', notice: 'Course could not be created.' }
       end
     end
   end
 
+  def help
+    @course = Course.find(params[:course_id])
```
jamox: use the prefore action (you'll see it after rebase)

jamox:
Tried merging and some tests were failing, maybe due to unsuccessful merge... 
Please rebase with the testmycode/tmc-server/tmc-teacher-tools so that it will merge clean.

PR #270, opened 2015-06-08
after migration, default organizations teachers are all tmc admins

FeisEater:


jamox:
It kinda hurts when we modify an existing migration, but I'll let it go as this is dev branch used by a small group of people.

PR #272, opened 2015-06-10
Rename tests to *_spec

jruo:
so we actually test them...


app/views/organizations/list_requests.html.erb
```diff
@@ -21,7 +21,7 @@
       <td><%= organization.created_at %></td>
       <td><%= link_to 'Show information', organization %></td>
       <td><%= link_to 'Accept', accept_organization_path(organization), method: :post, class: 'btn btn-success' %></td>
-      <td><%= link_to 'Reject', reject_reason_input_organization_path(organization), method: :post, class: 'btn btn-danger' %></td>
+      <td><%= link_to 'Reject', reject_reason_input_organization_path(organization), class: 'btn btn-danger' %></td>
```
jamox: Not a big fan of allowing changing state (i.e allowing to reject/accept org with a get request)

jamox:
This test keeps on failing after merge:
```
  1) Exercise should be submittable by non-administrators only if the deadline has not passed and the exercise is not hidden and is published
     Failure/Error: Unable to find matching line from backtrace
       expected `#<Exercise id: 11, name: "exercise8", created_at: "2015-06-10 21:55:51", updated_at: "2015-06-10 21:55:51", course_id: 10, publish_time: "2015-06-10 21:00:00", gdocs_sheet: "exercise11", hidden: false, returnable_forced: true, checksum: "", solution_visible_after: nil, has_tests: false, deadline_spec: nil, unlock_spec: nil, runtime_params: "[]", valgrind_strategy: nil, code_review_requests_enabled: false, run_tests_locally_action_enabled: true>.submittable_by?(#<User id: 6, login: "user5", password_hash: "f84c425d29339bd4121fb7e37f5748522b4a6dc71958a1f416...", created_at: "2015-06-10 21:55:51", updated_at: "2015-06-10 21:55:51", salt: "0d077478bdcfc22b65de225adf90a00e46ac4546bc70ecbf04...", administrator: false, email: "user5@example.com", legitimate_student: true>)` to return false, got true
     # /home/tmc/tmc-server/spec/models/exercise_spec.rb:182:in `block (2 levels) in <top (required)>'
```

app/views/organizations/list_requests.html.erb
jruo: The 'Reject' link does not actually reject the organization. It redirects to this page https://github.com/testmycode/tmc-server/blob/tmc-teacher-tools/app/views/organizations/reject_reason_input.html.erb where admin can give a reason for the rejection and then reject with a post request.

PR #273, opened 2015-06-10
Retrying pr 268 with cherry-picked commits

FeisEater:
Github still says it cannot automatically merge, yet when we merge this with our master (which is up to date with tmc-teacher-tools), no conflicts arise.


config/routes.rb
```diff
@@ -1,10 +1,12 @@
 TmcServer::Application.routes.draw do
+  get 'org/:organization_id/course_templates', to: 'course_templates#list_for_teachers', as: 'list_course_templates'
+  get 'org/:organization_id/course_templates/:id', to: 'course_templates#prepare_course', as: 'prepare_course'
```
jamox: Any reason for these manual matchers?

Termanty: No reason. I took matchers away and moved this routes inside organization.

Termanty: I was too fast. We need `as: 'prepare_course'` because rails can't create path-helper without it.

app/controllers/courses_controller.rb
```diff
@@ -132,6 +126,12 @@ def disable
     redirect_to(organization_course_path(@organization, @course), notice: 'Course was successfully disabled.')
   end
 
+
```
jamox: Two blank lines, remove one

PR #274, opened 2015-06-12
More course parameters

jruo:


app/controllers/courses_controller.rb
```diff
@@ -155,7 +168,7 @@ def save_unlocks
   private
 
   def course_params
-    params.permit(course: [:name, :source_url, :git_branch])
+    params.require(:course).permit(:name, :title, :description, :material_url, :source_url, :git_branch)
```
mpartel: For updates, we don't want to permit changing at least `name`.

app/views/courses/show.html.erb
```diff
@@ -1,7 +1,10 @@
 <div class="row-fluid">
   <div class="span12">
-    <h1><%= @course.name %></h1>
+    <h1><%= @course.title %></h1>
     <p><%= @course.description %></p>
+    <% unless @course.material_url.nil? || @course.material_url.empty? %>
```
mpartel: Use `blank?` instead of `nil? || empty?`

db/migrate/20150609083655_add_title_and_material_url_to_courses.rb
```diff
@@ -0,0 +1,11 @@
+class AddTitleAndMaterialUrlToCourses < ActiveRecord::Migration
+  def change
+    add_column :courses, :title, :string
+    add_column :courses, :material_url, :string
+    reversible do |dir|
+      dir.up do
+        Course.all.each { |c| c.title = c.name; c.save }
```
mpartel: `c.save!`

spec/controllers/courses_controller_spec.rb
```diff
@@ -229,6 +229,56 @@ def get_show_json(options = {}, parse_json = true)
...
+    describe 'when non-teacher attemps to update' do
+      before :each do
+        put :update, organization_id: @organization.to_param, id: @course.to_param, course: {title: 'newTitle', description: 'newDescription', material_url: 'newMaterial'}
+      end
+
+      it 'should respomd with 401' do
+        expect(response.code.to_i).to eq(401)
+      end
+
+      it 'shouldnt update' do
```
mpartel: `shouldn't` or `should not`

app/models/course.rb
```diff
@@ -17,6 +17,16 @@ class Course < ActiveRecord::Base
...
             }
 
+  validates :title,
+            presence: true,
+            length: { within: 1..40 }
+
+  validates :material_url,
+            format: {
+                without: /:/,
+                message: 'should not contain colon (:). Remove "http://" from the beginning if you have it'
```
mpartel: But what about https URLs?

I'd much rather require `/^https?:\/\//` in this field.

spec/features/teacher_disables_courses_spec.rb
```diff
@@ -8,7 +8,7 @@
     @teacher = FactoryGirl.create :user, password: '1234'
     @user = FactoryGirl.create :user, password: '1234'
     @organization = FactoryGirl.create :accepted_organization, slug: 'slug'
-    @course = FactoryGirl.create :course, name: 'test-course-1', organization: @organization
+    @course = FactoryGirl.create :course, name: 'test-course-1', title: 'test-course-1', organization: @organization
```
mpartel: (make title different from name here?)

spec/features/teacher_lists_own_organization_courses_spec.rb
```diff
@@ -9,9 +9,9 @@
     @user = FactoryGirl.create :user, password: 'foobar'
     Teachership.create! user: @teacher, organization: @organization
 
-    FactoryGirl.create :course, name: 'course_1', organization: @organization
-    FactoryGirl.create :course, name: 'course_2', organization: @organization
-    FactoryGirl.create :course, name: 'course_old', organization: @organization, hide_after: Time.now - 2.minutes
+    FactoryGirl.create :course, name: 'course_1', title: 'course_1', organization: @organization
+    FactoryGirl.create :course, name: 'course_2', title: 'course_2', organization: @organization
+    FactoryGirl.create :course, name: 'course_old', title: 'course_old', organization: @organization, hide_after: Time.now - 2.minutes
```
mpartel: (make titles different from names here?)

spec/integration/broken_utf8_spec.rb
```diff
@@ -7,7 +7,7 @@
     repo_path = Dir.pwd + '/remote_repo'
     create_bare_repo(repo_path)
     @organization = FactoryGirl.create(:accepted_organization, slug: 'slug')
-    @course = Course.create!(name: 'mycourse', source_backend: 'git', source_url: repo_path, organization_id: @organization.id)
+    @course = Course.create!(name: 'mycourse', title: 'mycourse', source_backend: 'git', source_url: repo_path, organization_id: @organization.id)
```
mpartel: (make title different from name here?)

spec/models/course_spec.rb
```diff
@@ -178,11 +179,25 @@
...
+    it 'requires title to be reasonably short' do
+      should_be_invalid_params(valid_params.merge(title: 'a' * 41))
+    end
+
     it 'requires a remote repo url' do
       should_be_invalid_params(valid_params.merge(source_url: nil))
       should_be_invalid_params(valid_params.merge(source_url: ''))
     end
 
+    it 'doesnt allow colon in material url' do
```
mpartel: `does not`

spec/support/submission_test_setup.rb
```diff
@@ -25,7 +25,7 @@ def initialize(options = {})
     @repo_path = 'remote_repo'
     create_bare_repo(@repo_path)
 
-    @course = Course.create!(name: course_name, source_backend: 'git', source_url: @repo_path, organization: organization)
+    @course = Course.create!(name: course_name, title: course_name, source_backend: 'git', source_url: @repo_path, organization: organization)
```
mpartel: How about `title: course_name.humanize`?

app/controllers/courses_controller.rb
```diff
@@ -5,7 +5,7 @@
 
 class CoursesController < ApplicationController
   before_action :set_organization
-  before_action :set_course, only: [:show, :refresh, :manage_deadlines, :save_deadlines, :enable, :disable, :manage_unlocks, :save_unlocks]
+  before_action :set_course, only: [:show, :edit, :update, :refresh, :manage_deadlines, :save_deadlines, :enable, :disable, :manage_unlocks, :save_unlocks ]
```
jamox: Maybe we should sort these, or use the `except` instead of `only`

app/models/course.rb
jamox: +1

app/models/course.rb
```diff
@@ -89,7 +99,7 @@ def options=(new_options)
     self.hidden = !!new_options['hidden']
     self.spreadsheet_key = new_options['spreadsheet_key']
 
-    self.description = new_options['description']
+    # self.description = new_options['description'] # disabled for now, description read from course template or db
```
jamox: why not just rm?

app/controllers/courses_controller.rb
jamox: ping^
this has more consequences than one might think (e.g. how files are stored by clients)

app/models/course.rb
jamox: yeah, this doesn't sound like a reasonable request?
And since we can always preprocess the input; its more user friendly if the user can just copy/paste the url from his/her browser.

app/views/courses/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Edit course</h1>
+
+<%= form_for(@course, url: organization_course_path) do |f| %>
+  <%= render 'shared/error_messages', target: @course %>
+
+  <%= labeled_field("Course title", f.text_field(:title)) %>
+  <%= labeled_field("Course description (Optional)", f.text_area(:description)) %>
+  <%= labeled_field("Material URL (Optional)", f.text_field(:material_url)) %>
```
jamox: I'd rather have the optionality to be a visual cue

app/views/courses/help.html.erb
```diff
@@ -18,3 +18,8 @@
 
 <p><%= link_to organization_course_url(@organization, @course), organization_course_url(@organization, @course) %></p>
 <p>This url can be used with Netbeans plugin.</p>
+
+<% unless @course.material_url.nil? || @course.material_url.empty? %>
```
jamox: .blank? checks for nil and if it has any content

app/controllers/courses_controller.rb
FeisEater: We will probably rewrite this line at some later pr, when we will have most of controller methods implemented and merged

spec/support/submission_test_setup.rb
FeisEater: This change would require changes in several other test files, if they expect to click a link with course's name on it. We changed it so course's title is shown instead.

FeisEater:
Fixed based on your comments, except for some parts. Write back if you insist on changing these parts as well.

PR #275, opened 2015-06-12
Soft deadlines

FeisEater:


db/schema.rb
```diff
@@ -95,6 +95,7 @@
     t.string   "valgrind_strategy"
     t.boolean  "code_review_requests_enabled",     default: false, null: false
     t.boolean  "run_tests_locally_action_enabled", default: true,  null: false
+    t.text     "soft_deadline_spec"
```
jamox: Wouldn't string suffice?

jamox:
Maybe we should also be updating the [usermanual](https://github.com/testmycode/tmc-server/blob/master/spec/usermanual/instructors.html.erb)

db/schema.rb
jruo: Probably yes, but the other _specs are text too. Should I change them?

jruo:
We are updating the manual right now. It will be in its own branch.

jruo:
Arto thought that students don't need to know about hard deadlines, but I changed it that it now shows hard deadline if soft deadline is not specified

db/schema.rb
jamox: ok, we can leave it as it is.

jamox:
One question remains, what happens to current courses without soft deadline spec?

jruo:
They should work exaclty the same. Soft deadline specs can be null, and they don't change the way the current system works.

app/models/exercise_group.rb
```diff
@@ -75,4 +80,17 @@ def group_unlock_conditions=(unlock_conditions)
...
+  private
+
+  def group_deadline(method)
+    exercises(false).map { |n| n.send(method) }.first
+  end
+
+  def set_group_deadline(method, deadline)
+    exercises(false).each do |e|
+      e.send(method, deadline)
+      e.save
```
jamox: save!

spec/models/exercise_spec.rb
```diff
@@ -99,6 +99,14 @@ def set_deadline(ex, t)
     end
   end
 
+  def set_soft_deadline(ex, t)
+    if t.is_a? Array
+      ex.soft_deadline_spec = t.to_json
+    else
+      ex.soft_deadline_spec = [t.to_s].to_json
+    end
+  end
```
jamox: I don't like the .is_a? -- could we maybe survive with flatten?
And what t is if it is not an array

jruo:
Updated

app/controllers/courses_controller.rb
```diff
@@ -105,8 +105,10 @@ def save_deadlines
 
     groups = group_params
     groups.each do |name, deadlines|
-      json_array = [deadlines[:static], deadlines[:unlock]].to_json
-      @course.exercise_group_by_name(name).group_deadline=(json_array)
+      soft_deadlines = [deadlines[:soft][:static], deadlines[:soft][:unlock]].to_json
+      hard_deadlines = [deadlines[:hard][:static], deadlines[:hard][:unlock]].to_json
```
jamox: should we somehow validate the input/ is it to be validated at some point?

jruo: When we pass the input to the exercises, they try to parse it into a spec. If it fails, they raise an invalidsyntaxerror which is caught at the end of this method.

PR #277, opened 2015-06-15
course notification emailing rights for teacher

jruo:


app/controllers/course_notifications_controller.rb
```diff
@@ -47,7 +47,7 @@ def course_notification_params
   end
 
   def auth
-    authorize! :email, CourseNotification
+    authorize! :teach, @organization
```
jamox: I'd prefer to have `authorize! :send_email, CourseNotification, @organization`

mpartel: Agreed (I weakly prefer the 2nd suggestion.)

Looks good otherwise.

PR #278, opened 2015-06-15
Course assistants

Termanty:


app/views/assistants/index.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <tbody>
+  <% @assistants.each do |assistant| %>
+    <tr>
+      <td><%= assistant.username %></td>
+    </tr>
+  <% end %>
+  </tbody>
+</table>
+
+<%= link_to 'Add a new assistant', new_organization_course_assistant_path(@course.organization, @course), class: "btn btn-mini btn-warning" %>
```
jamox: this could also be a embedded form?

app/models/course.rb
```diff
@@ -42,6 +44,7 @@ def destroy
     # Even self.association.delete_all first does a SELECT.
     # This relies on the database to cascade deletes.
     ActiveRecord::Base.connection.execute("DELETE FROM courses WHERE id = #{id}")
+    assistantships.each { |a| a.destroy! } # apparently this is not performed automatically with optimized destroy
```
jamox: The course should really never get deleted though...
Nice to have it anyway

app/models/course.rb
```diff
@@ -290,6 +293,10 @@ def initially_refreshed?
     refreshed_at.nil?
   end
 
+  def assistant?(user)
+    assistants.include? user
```
jamox: I'm wondering whether this is basically `(select * from assistants where course_id=...).to_a.include?` or if this is performed with actual sql querty. with bigger courses this might become really ineficient

app/models/user.rb
```diff
@@ -121,6 +123,10 @@ def teacher?(organization)
     organizations.include? organization
   end
 
+  def assistant?(course)
+    assisted_courses.include? course
```
jamox: see comment abobe

app/views/assistants/index.html.erb
```diff
@@ -0,0 +1,19 @@
+<h1>Assistants in <%= @course.title %></h1>
+
+<table>
+  <thead>
+  <tr>
+    <th>Name</th>
+  </tr>
```
jamox: The `tr` should be indented

jamox:
This is a step towards adding more controllable permissions; though always adding more and more relations doesn't really scale.

However I consider this to be acceptable step towards goal of having more better control of permissions.
@mpartel do you have strong opinions regarding this step towards 'better permissions'?

FeisEater:
Fixed assistant? methods, should be doing only one SQL query

app/controllers/courses_controller.rb
```diff
@@ -121,7 +121,7 @@ def save_deadlines
   end
 
   def enable
-    authorize! :teach, @organization
+    authorize! :teach, @organization # should assistants be able to enable/disable?
```
mpartel: There is probably not enough time in this project but ideally we'd have more fine-grained abilities and allow organization admins to delegate them selectively.

app/models/assistantship.rb
```diff
@@ -0,0 +1,9 @@
+class Assistantship < ActiveRecord::Base
+  belongs_to :user
+  belongs_to :course
+
+  validates :user,
+            presence: { message: 'does not exist' },
+            uniqueness: { scope: :course, message: 'is already an assistant for this course' }
+  validates :course, presence: true
+end
```
mpartel: This is fine for now but a more generic solution would be to have a table of `(user_id, course_id, role or permission)`. Now `Assistantship` and `Tearchership` do pretty much the same thing.

app/models/course.rb
```diff
@@ -290,6 +293,10 @@ def initially_refreshed?
     refreshed_at.nil?
   end
```
mpartel: This has only been used once in production, which we later predictably regretted. There are probably a whiole bunch of things that this code leaves behind as junk. Feel free to just remove this action (and the corresponding link in the UI).

db/migrate/20150611073002_create_assistantships.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAssistantships < ActiveRecord::Migration
+  def change
+    create_table :assistantships do |t|
+      t.references :user, index: true
+      t.references :course, index: true
```
mpartel: I think there should be just one index of `(user_id, course_id)` and it should be unique.

Also, please add a database-level foreign key to both course and user with `:dependent => :delete` for both.

Please do all this for teacherships too.

spec/models/assistantship_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    course1 = FactoryGirl.create :course
+    course2 = FactoryGirl.create :course
+    Assistantship.create! user: user1, course: course1
+    expect(user1.assisted_courses).to eq([course1])
+    expect(user2.assisted_courses).to eq([])
+    expect(course1.assistants).to eq([user1])
+    expect(course2.assistants).to eq([])
+  end
+
+  it 'cant create assistantship if course is non-existant' do
```
mpartel: "can't" or "cannot", same below.

app/controllers/courses_controller.rb
mpartel: As I notice @jamox already said above..

I agree, OK to leave for now but should definitely refactor if we ever want to develop permissions further.

app/views/assistants/index.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  </thead>
+
+  <tbody>
+  <% @assistants.each do |assistant| %>
+    <tr>
+      <td><%= assistant.username %></td>
+    </tr>
+  <% end %>
+  </tbody>
+</table>
```
jamox: A list might suffice...

app/views/assistants/index.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <thead>
+    <tr>
+      <th>Name</th>
+    </tr>
+  </thead>
+
+  <tbody>
+  <% @assistants.each do |assistant| %>
+    <tr>
+      <td><%= assistant.username %></td>
```
jamox: Maybe link to the user

jamox:
LGTM except for the few comments

db/migrate/20150611073002_create_assistantships.rb
FeisEater: I'll do this for teachers in another branch

app/views/assistants/index.html.erb
jsopakar: Little bit permission tuning needed for this to work well, will be done in bigger context in later PR, marked as own story/issue (https://github.com/TMC-teacher-tools/tmc-server/issues/75)

app/views/assistants/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1>Assistants in <%= @course.title %></h1>
+
+<ul>
+  <% @assistants.each do |assistant| %>
+    <li><%= link_to assistant.username, participant_path(assistant) %></li>
```
jamox: Note for myself: shouldn't this be login, not username...

jamox: ok, so the login is the field, but we have a hax alias for it. nice

FeisEater:
Should be able to auto-merge, and tests should pass if internet is on

PR #280, opened 2015-06-17
Teacher lists student emails

FeisEater:


app/controllers/courses_controller.rb
```diff
@@ -152,6 +152,22 @@ def save_unlocks
     redirect_to manage_unlocks_organization_course_path(@organization, @course), alert: e.to_s
   end
 
+  def student_emails
```
mpartel: This functionality seems to overlap with `EmailsController` and `ParticipantsController`. I think we should avoid duplicating the functionality and refactor the former to be suitable for what you need instead.

app/controllers/courses_controller.rb
```diff
@@ -152,6 +152,22 @@ def save_unlocks
...
+    authorize! :teach, @organization
+    add_course_breadcrumb
+    add_breadcrumb('Students', students_organization_course_path(@organization, @course))
+    @students = []
+    @course.submissions.each do |s|
```
mpartel: This loop will be quite slow. Maybe use `User.course_students(@course)`.

app/views/courses/student_emails.html.erb
```diff
@@ -0,0 +1,26 @@
...
+      Email
+    </th>
+  </tr>
+  <% @students.each do |student| %>
+    <tr>
+      <td>
+        <%= student.username %>
+      </td>
+      <td>
+        <a href="mailto:<%= student.email %>"><%= student.email %></a>
```
jamox: Use rather: http://apidock.com/rails/ActionView/Helpers/UrlHelper/mail_to

app/controllers/courses_controller.rb
```diff
@@ -152,6 +152,22 @@ def save_unlocks
...
+      @students << s.user unless @students.include? s.user
+    end
+
+    respond_to do |format|
+      format.html
+      format.text { render text: @students.map { |s| "#{s.email}" }.join("\n") }
+      format.csv { render text: "Username,Email\n" + @students.map { |s| "#{s.username},#{s.email}"}.join("\n") }
+    end
+  end
+
```
jamox: Couldn't we easily do this with participants list -- have it courses/:id/participants, with all participants for this course

config/routes.rb
```diff
@@ -25,6 +25,7 @@
         post 'disable'
         get 'manage_unlocks'
         post 'save_unlocks'
+        get 'students', to: 'courses#student_emails'
```
jamox: could have the participants nested here... see comment above

jruo:
Updated. The ParticipantsController seemed a bit too robust for a simple email list, so I used EmailsController instead.

PR #283, opened 2015-06-18
Teachers can create courses from their own repos

jruo:


jamox:
Please include flag to site.yml to enable/disable this feature. (for them to be able to provide own git repo)

mpartel:
I'm not sure the flag is a good idea.

As we discussed offline and as I discussed with @jamox later, it'd be best to migrate all courses on top of templates and not have courses without templates at all. This would avoid lots of complexity. We'd also really like courses using the same template to use the same git repo before merging this.

jamox:
The flag so that we won't accept unsafe git sources before clone/stub and model solution creation etc is sandboxed

app/controllers/course_templates_controller.rb
```diff
@@ -52,7 +52,8 @@ def prepare_course
                          title: @course_template.title,
                          description: @course_template.description,
                          material_url: @course_template.material_url,
-                         source_url: @course_template.source_url
+                         source_url: @course_template.source_url,
+                         course_template_id: @course_template.id
```
jamox: maybe add brackets for the Course.new, even though it's not explicitly required

app/views/course_templates/list_for_teachers.html.erb
```diff
@@ -1,4 +1,6 @@
-<h1>Listing course templates</h1>
+<h1>New Course</h1>
+
+<h3>Create course from templates</h3>
```
jamox: from template -- one course is created from one template, right?

mpartel:
LGTM

app/views/course_templates/list_for_teachers.html.erb
Termanty: That will be implemented in later PR

spec/support/integration_test_actions.rb
```diff
@@ -112,4 +115,8 @@ def trim_image_edges(path)
     system!(cmd)
     system!(cmd2)
   end
+
+  def check_custom_enabled
+    pending 'Enable custom repositories from site.defaults.yml' unless SiteSetting.value('enable_custom_repositories')
```
jamox: not sure if such test is needed? (do we have similar for other values)

jruo: Not really needed, but otherwise the tests will simply fail if the setting is disabled. I can remove it if you want.

PR #284, opened 2015-06-18
JSON page that shows submissions

jruo:


app/controllers/courses_controller.rb
```diff
@@ -70,6 +70,22 @@ def show_json
     render json: data.to_json
   end
 
+  # This is for analysis and better points from exercise program, which will be done one day.
+  def submissions_data
```
jamox: Name can be too easily confused with [submission_data.rb](https://github.com/testmycode/tmc-server/blob/master/app/models/submission_data.rb)

lib/exercise_list.rb
```diff
@@ -0,0 +1,25 @@
...
+        id: exercise.id,
+        exercise_name: exercise.name,
+        course_id: exercise.course_id,
+        available_points: AvailablePointList.new.available_point_list(exercise.available_points),
+        created_at: exercise.created_at,
+        updated_at: exercise.updated_at,
+        publish_time: exercise.publish_time,
+        hidden: exercise.hidden,
+        solution_visible_after: exercise.solution_visible_after,
+        has_tests: exercise.has_tests,
```
jamox: No need for this field yet

lib/exercise_list.rb
```diff
@@ -0,0 +1,25 @@
...
+    {
+        id: exercise.id,
+        exercise_name: exercise.name,
+        course_id: exercise.course_id,
+        available_points: AvailablePointList.new.available_point_list(exercise.available_points),
+        created_at: exercise.created_at,
+        updated_at: exercise.updated_at,
+        publish_time: exercise.publish_time,
+        hidden: exercise.hidden,
+        solution_visible_after: exercise.solution_visible_after,
```
jamox: why?

lib/exercise_list.rb
```diff
@@ -0,0 +1,25 @@
...
+        has_tests: exercise.has_tests,
+        deadline_spec: exercise.deadline_spec,
+        soft_deadline_spec: exercise.soft_deadline_spec,
+        unlock_spec: exercise.unlock_spec,
+        code_review_requests_enabled: exercise.code_review_requests_enabled,
+        run_tests_locally_action_enabled: exercise.run_tests_locally_action_enabled
```
jamox: I don't see need for all of these fields? -- I'd avoid exposing data which is meant to be used internally without a clear need/external consumer for that information.

app/controllers/courses_controller.rb
```diff
@@ -70,6 +70,22 @@ def show_json
...
+    authorize! :teach, @organization
+
+    submissions = @course.submissions.order('created_at DESC')
+    exercises = @course.exercises
+    data = {
+        api_version:                      ApiVersion::API_VERSION,
+        course_name:                      @course.name,
+        course_id:                        @course.id,
+        exercises:                        ExerciseList.new.exercise_list(exercises),
+        submissions:                      SubmissionList.new(current_user, view_context).submission_list_for_analysis(submissions)
```
mpartel: Too much indentation.

lib/available_point_list.rb
```diff
@@ -0,0 +1,14 @@
...
+  def available_point_list(available_points)
+    available_points.map { |ap| available_point_data(ap) }
+  end
+
+  def available_point_data(available_point)
+    {
+        id: available_point.id,
+        name: available_point.name,
+        exercise_id: available_point.exercise_id,
+        requires_review: available_point.requires_review
```
mpartel: Too much indentation.

jamox:
I don't see clear use for this; if for a system which uses points/amount of submissions, what we should be doing is more fetching points for given user, or somehow crawl snapshot once and then use a hook in when submission is graded, to notify the app of the received submission.

And this would be a huge issue for courses where we have say 200 000 submissions.

And we already have api `courses.json` and `exercises/:id.json` which povides more or less all the really needed fields...

PR #286, opened 2015-06-22
Admin features to teacher

Termanty:
We would like to know if permissions are now going to right direction.
PR - Course assistants #278 should be merged before this.


app/models/user.rb
```diff
@@ -117,10 +119,59 @@ def <=>(other)
...
+  def visible_to_teacher?(teacher)
+    teacher.organizations.each do |org|
+      return true if self.student_in_organization?(org)
+    end
+    false
+  end
+
+  def visible_to_assistant?(assistant)
+    assistant.assisted_courses.each do |c|
+      return true if self.student_in_course?(c)
```
jamox: This seems to be creating n queries to db?

jsopakar: As discusses bit more in irc, simplified this part a bit, but visibility stays bit heavy operation for now, and shouldn't be used in looping for views' list generation. Relevant comments added.

jsopakar:
I merged current upstream to this because of conflicting changes, so this pr should should be mergeable now, if no new problems found.

PR #288, opened 2015-06-24
Update breadcrumbs and site flow

jsopakar:
Organization index page now handles showing courses, courses index just redirects to this, to simplify the flow.

### DIFF
--- a/app/controllers/breadcrumb_helpers.rb
+++ b/app/controllers/breadcrumb_helpers.rb
@@ -3,7 +3,12 @@
 #
 # See https://github.com/weppos/breadcrumbs_on_rails
 module BreadcrumbHelpers # Included in ApplicationController
+  def add_organization_breadcrumb
+    add_breadcrumb "Organization #{@organization.name}", organization_path(@organization)
+  end
+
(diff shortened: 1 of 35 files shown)


app/controllers/breadcrumb_helpers.rb
```diff
@@ -3,7 +3,12 @@
 #
 # See https://github.com/weppos/breadcrumbs_on_rails
 module BreadcrumbHelpers # Included in ApplicationController
+  def add_organization_breadcrumb
+    add_breadcrumb "Organization #{@organization.name}", organization_path(@organization)
```
mpartel: This could have `organization = @organization || @course.organization` which might help avoid some `@organization = @course.organization` statements elsewhere. But do this only if you agree it's an improvement, I'm not entirely sure.

mpartel:
LGTM, nice cleanup!

app/controllers/organizations_controller.rb
```diff
@@ -78,7 +88,7 @@ def reject
       @organization.save
```
jamox: not checking return value

PR #289, opened 2015-06-24
Course template from factorygirl uses local repo instead of cloning from internet

jsopakar:


jamox:
One test in user manual was failing, when no internet was available. Please fix in a new PR

PR #291, opened 2015-06-25
Teachers can remove other teachers and assistants

jsopakar:


app/controllers/assistants_controller.rb
```diff
@@ -21,6 +21,13 @@ def create
     end
   end
 
+  def destroy
+    authorize! :remove_assistant, @course
+    @assistantship = Assistantship.find(params[:id])
+    @assistantship.destroy!
+    redirect_to organization_course_assistants_path ,notice: 'Assistant removed from course'
```
mpartel: (misplaced comma)

app/models/ability.rb
```diff
@@ -92,11 +92,19 @@ def initialize(user)
...
+      end
+
       cannot :teach, Course
       can :teach, Course do |c|
         return false if c.organization.rejected?
         c.organization.teacher?(user) || c.assistant?(user)
       end
+
+      can :remove_assistant, Course do |c|
+        c.organization.teacher?(user)
```
mpartel: No `!rejected?` and `!acceptance_pending?` checks here?

app/views/assistants/_form.html.erb
```diff
@@ -1,10 +1,9 @@
-<h1>Add a new assistant to <%= @course.title %></h1>
+<h4>Add a new assistant (give username)</h4>
```
mpartel: Why remove the course title?

mpartel:
LGTM other than the above.

app/views/assistants/_form.html.erb
Termanty: Visual reason - for me page just looks better without it.

mpartel: Ah, never mind, I was thinking of the form as an entire page but it isn't. Ignore my silliness.

PR #293, opened 2015-06-26
More permission fixes

jruo:


app/models/ability.rb
```diff
@@ -141,6 +141,30 @@ def initialize(user)
...
+        can? :teach, c
+      end
+
+      can :manage_exercises, Course do |c|
+        can? :teach, c.organization
+      end
+
+      can :edit_course_paramaters, Course do |c|
+        can? :teach, c.organization
+      end
```
jamox: I'm wondering if these should be ordered by some logic, alphabetically or maybe grouped by logical uses

PR #294, opened 2015-06-29
Organization hiding

jsopakar:
Functionality for teachers (and admins) to toggle organization visibility.


app/controllers/organizations_controller.rb
```diff
@@ -82,6 +76,13 @@ def reject
     end
   end
 
+  def toggle_hidden
+    authorize! :toggle_hidden, @organization
```
jamox: toggle_visiblility?

Termanty: sounds much better!!

PR #295, opened 2015-06-29
Bind repo dir to course template and make template refresh to handle refreshing all valid courses

jsopakar:
Courses created from same course template will use same repo directory, instead of all creating their own.

This PR is meant for comments about changes, as the author can't participate anymore, and responsibility for this carries on. Many tests failed still, but I have to take look at those closer soon.

So this is not meant to be ready for mergeing yet!


app/controllers/course_templates_controller.rb
```diff
@@ -52,7 +52,9 @@ def prepare_course
                          title: @course_template.title,
                          description: @course_template.description,
                          material_url: @course_template.material_url,
-                         source_url: @course_template.source_url
+                         source_url: @course_template.source_url,
+                         course_template_id: @course_template.id,
+                         cache_version: @course_template.cache_version
```
jamox: maybe have also the last line ending with `,` so it would be easier to reorder/add new lines

app/models/course.rb
```diff
@@ -305,8 +318,22 @@ def material_url=(material)
...
+  def custom?
+    course_template_id.nil?
+  end
+
   private
 
+  def check_custom_course_name_not_taken
+    if custom?
+      custom_courses = Course.where(course_template: nil)
+      custom_courses = custom_courses.where('id <> ?', self.id) unless self.id.nil?
```
jamox: not big fan of this syntax, maybe there is nicer way to do it.

app/models/course_template.rb
```diff
@@ -37,4 +43,35 @@ def valid_source_url?
...
+    "#{Course.cache_root}/#{name}-#{cache_version}"
+  end
+
+  def refresh
+    firstcourse = true
+    courses.each do |c|
+      c.refresh no_directory_changes: !firstcourse
+      reload
+      firstcourse = false
+    end
```
jamox: this looks like a hack...

app/models/course_template.rb
```diff
@@ -37,4 +43,35 @@ def valid_source_url?
...
+  end
+
+  def cache_exists?
+    File.exist?(cache_path)
+  end
+
+  private
+
+  def delete_courses_if_custom_courses_disabled
+    courses.each { |c| c.destroy! } if !custom_courses_enabled?
```
jamox: Not sure if we ever wan't to do this...

app/models/course_template.rb
jsopakar: Hmm, when thinking of this, cleaner way could be that every course would be based on course template (even the old ones and single custom courses). Then all the refresh logic for repo and file handling could be made always for template only. 

But I think it takes quite big refactoring of whole refresh system... Not sure if it would be wiser to implement later on different task. Even though it was discussed earlier too elsewhere.

app/controllers/course_templates_controller.rb
```diff
@@ -76,6 +77,16 @@ def toggle_hidden
     redirect_to course_templates_path, notice: 'Course templates hidden status changed'
   end
 
+  def refresh
+    notice = "All good"
+    begin
+      @course_template.refresh
+    rescue CourseRefresher::Failure => e
+      notice = "Something fucked up"
```
jamox: include the exception + stacktrace

app/models/ability.rb
```diff
@@ -38,6 +38,11 @@ def initialize(user)
       can :create, Course do |c|
         can? :teach, c.organization
       end
+      can :refresh, Course do |c|
+        c.taught_by?(user) &&
+            c.custom? && # user can only refresh his/her custom course.
+            SiteSetting.value('enable_custom_repositories') # and only if custom repositories are allowed
```
jamox: this was removed/modified to check whether user is admin...

app/models/course_template.rb
```diff
@@ -26,6 +27,9 @@ class CourseTemplate < ActiveRecord::Base
   scope :not_hidden, -> { where(hidden: false) }
   scope :available, -> { not_expired.not_hidden }
 
+  after_destroy :delete_courses_if_custom_courses_disabled
+  after_destroy :delete_cache
```
jamox: maybe not allow full deletion; maybe just add deleted_at: datetime and if set, consider it to be deleted.. maybe

app/models/course_template.rb
jamox: Yeah - I would consider it to be acceptable if we force each course to depend on course_template. but yes as long as it gets done, it doesn't need to be part of this PR.

db/migrate/20150518095422_create_course_templates.rb
```diff
@@ -6,6 +6,7 @@ def change
       t.string :description
       t.string :material_url
       t.string :source_url
+      t.integer  "cache_version",   default: 0,     null: false
```
jamox: t.integer :cache_version......

lib/course_refresher.rb
```diff
@@ -129,12 +130,12 @@ def refresh_course(course)
         rescue StandardError, ScriptError # Some YAML parsers throw ScriptError on syntax errors
           @report.errors << $!.message + "\n" + $!.backtrace.join("\n")
           # Delete the new cache we were working on
-          FileUtils.rm_rf(@course.cache_path)
+          FileUtils.rm_rf(@course.cache_path)                     unless options[:no_directory_changes]
```
jamox: I'd consider just having one space before the unless - this looks like too much indentation and adds overhead to change it if there is more steps added to the refresh process

app/models/course_template.rb
jsopakar: I'm thinking that should course template deleting be allowed at all, if there are any courses made from that template? There's anyway hidden field plus expiredate for course templates already, to let old templates pass away.

lib/course_refresher.rb
jsopakar: Was this meant for all same kind of unless rows, or just for the quoted one? Now on the list above that it looks quite clear imo, but can take idents away of course if wanted.

app/models/course.rb
jsopakar: I simplified this a bit

jsopakar:
This should be now ready to merge from functionality point of view. Some comments about identitation, and delete logic are still unanswered though.

jsopakar:
Reports from template refresh are also printed in the same kind of style as before from single course refresh. This propably is enough for now, but in the future when there can be tens of courses from same template, this can be bit messy. Not sure if we have to think better raporting then.

app/controllers/course_templates_controller.rb
```diff
@@ -1,11 +1,15 @@
 class CourseTemplatesController < ApplicationController
-  before_action :set_course_template, only: [:edit, :update, :destroy, :prepare_course, :toggle_hidden]
+  before_action :set_course_template, only: [:edit, :update, :destroy, :prepare_course, :toggle_hidden, :refresh]
```
jamox: maybe switch to except; this list is getting quite long

app/controllers/courses_controller.rb
```diff
@@ -205,7 +209,11 @@ def course_params_for_create
   end
 
   def course_params
-    params.require(:course).permit(:title, :description, :material_url, :source_url, :git_branch)
+    if (@course.custom?)
```
jamox: `if` w/o brackets

app/models/ability.rb
```diff
@@ -8,7 +8,10 @@ def initialize(user)
     if user.administrator?
       can :manage, :all
       can :create, Course
-      can :refresh, Course
+      cannot :refresh, Course
```
jamox: I think admins should be allowed to refresh the course just in case of we need to do some sort of maintenance etc.

jsopakar: Hmm. At this point it can be problematic for non-custom course. In those cases, straight refresh for the course can cause other courses from same template be non-synced. If needed, can create link for template refresh-instead actually..

This will be changed to more clear solution later, after all courses are migrated to have template.

jsopakar:
Fixed those minor requests, permission thing still as it was, with comment above, plus added information of courses template, including link to template list (where admin can do the refresh). Linking is not perfect, as templates didn't have any show-controller, but I think this is reasonable for admin case.

jamox:
The course template functionality was merged via another PR.

PR #296, opened 2015-06-30
Admin can create course from own repo

Termanty:


app/views/courses/show.html.erb
```diff
@@ -26,16 +26,17 @@
...
-            <% end %>
+              <% github_url = github_repo_url_to_project_page_url(@course.source_url) %>
+              <% if github_url %>
+                <li>Github URL: <%= link_to github_url, github_url %></li>
+              <% end %>
+              <li>Git branch: <%= @course.git_branch %></li>
+              <% gitrev = @course.git_revision %>
+              <% if gitrev %>
+                <li>Git revision: <%= gitrev %></li>
+              <% end %>
```
jamox: is this change for indentation correct?

db/migrate/20150616081429_add_course_template_id_to_course.rb
```diff
@@ -0,0 +1,5 @@
+class AddCourseTemplateIdToCourse < ActiveRecord::Migration
+  def change
+    add_column :courses, :course_template_id, :integer
```
jamox: Should we maybe have index for this field?

app/views/courses/show.html.erb
Termanty: This is already so awful that soon I will refactor this file.

PR #297, opened 2015-07-01
Teacher can give multiple unlock conditions

Termanty:
Simpler way to create multiple unlock condition from UI.


app/views/courses/manage_unlocks.erb
```diff
@@ -1,43 +1,69 @@
-<h1>Edit unlock dates</h1>
```
jsopakar: Filename missing html part?

jruo: Fixed

app/controllers/courses_controller.rb
```diff
@@ -180,7 +180,7 @@ def save_unlocks
     groups = group_params
     groups.each do |name, conditions|
       array = []
-      conditions.each { |k, v| array << v }
+      conditions.each { |k, v| array << v unless v.blank? }
```
jamox: I'm wondering if we have better way for doing this; `.flatten!` and `.uniq!` might be helpful.

jruo:
Updated

jruo:
Added button to remove conditions (https://github.com/TMC-teacher-tools/tmc-server/issues/95)

jamox:
Merged manually to current master.

PR #298, opened 2015-07-02
Logo for organization

jsopakar:


app/views/organizations/show.html.erb
```diff
@@ -1,4 +1,13 @@
-<h1><%= @organization.name %></h1>
+
+
```
jsopakar: 2 empty rows?

spec/usermanual/teachers.html.erb
```diff
@@ -93,6 +93,7 @@ require 'cgi' # for escapeHTML
     <ul>
       <li><b>Name</b> of the organization that will be seen in the app</li>
       <li><b>Description</b> about your organization, seen in this organization's page</li>
+      <li><b>Logo</b> for your organization, seen in this organization's page (you can change this and other fields above later)</li>
```
jsopakar: Was there some requirements for the logo, like minimum size? If not, good, if there is, would be good to tell them here.

Termanty: Any size is ok. Paperclip will automagic way scale picture 
to right size.

jamox: Though we don't really want to upscale the image...

Termanty: Current size of the logo is 100*100. I think most of the logos are bigger than that. Should we still restrict acceptance of images at least that 100*100 size?

jamox: At max maybe add note of the recommended size to the teachers.html.erb...

jamox:
LGTM

jsopakar:
Closed, implemented in new pr, #321

PR #302, opened 2015-07-09
Submission result visibility

jsopakar:
Especially for machine test purposes, added ability to hide submission results for course. Teacher can toggle that on/off. If results are hidded, student can't see anywhere, if the tests passed or not.

### DIFF
--- a/app/assets/stylesheets/main.css
+++ b/app/assets/stylesheets/main.css
@@ -10,6 +10,10 @@
   color: green;
 }
 
+.h_i_d_d_e_n {  /* Submision status :hidden is change for css to be 'h_i_d_d_e_n' so that we can see it (: */
+    color: orange;
+}
+
 /* Basic commons */
 
 .small{
--- a/app/assets/stylesheets/main.css
+++ b/app/assets/stylesheets/main.css
@@ -10,8 +10,8 @@
   color: green;
 }
 
-.h_i_d_d_e_n {  /* Submision status :hidden is change for css to be 'h_i_d_d_e_n' so that we can see it (: */
-    color: orange;
+.hidden_status {
+  color: orange;
 }
 
 /* Basic commons */
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -200,9 +200,8 @@ class CoursesController < ApplicationController
 
   def toggle_submission_result_visibility
     authorize! :toggle_submission_result_visibility, @course
-    @course.hide_submission_result = !@course.hide_submission_result
-    @course.save!
-    redirect_to organization_course_path(@organization, @course), notice: "Submission results are now #{@course.hide_submission_result ? 'hidden' : 'visible'}"
+    @course.toggle_submission_result_visiblity
+    redirect_to organization_course_path(@organization, @course), notice: "Submission results are now #{@course.hide_submission_results ? 'hidden' : 'visible'}"
   end
 
   private
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -198,6 +198,13 @@ class CoursesController < ApplicationController
     @exercises_id_map = @exercises.map { |e| [e.id, e] }.to_h
   end
 
+  def toggle_submission_result_visibility
+    authorize! :toggle_submission_result_visibility, @course
+    @course.hide_submission_result = !@course.hide_submission_result
+    @course.save!
+    redirect_to organization_course_path(@organization, @course), notice: "Submission results are now #{@course.hide_submission_result ? 'hidden' : 'visible'}"
+  end
+
   private
 
   def course_params_for_create
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -45,14 +45,14 @@ class SubmissionsController < ApplicationController
           user_id: @submission.user_id,
           course: @course.name,
           exercise_name: @submission.exercise.name,
-          status: @submission.status,
+          status: @submission.status(current_user),
           points: @submission.points_list,
           processing_time: @submission.processing_time,
           message_for_paste: @submission.message_for_paste,
           missing_review_points: @exercise.missing_review_points_for(@submission.user)
         }
         output = output.merge(
-          case @submission.status
+          case @submission.status(current_user)
           when :processing then {
             submissions_before_this: @submission.unprocessed_submissions_before_this,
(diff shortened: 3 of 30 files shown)


app/assets/stylesheets/main.css
```diff
@@ -10,6 +10,10 @@
   color: green;
 }
 
+.h_i_d_d_e_n {  /* Submision status :hidden is change for css to be 'h_i_d_d_e_n' so that we can see it (: */
+    color: orange;
```
jsopakar: too much ident?

app/models/ability.rb
```diff
@@ -60,6 +60,14 @@ def initialize(user)
         can? :teach, sub.course
       end
 
+      can :download, Submission do |sub|
+        !sub.course.hide_submission_result? && can?(:read, sub)
+      end
+
+      can :read_result, Submission do |sub|
+        !sub.course.hide_submission_result? || can?(:teach, sub.course)
```
jsopakar: using brackets with can? clause is different than elsewhere in file. not sure which is best, but either should use this, or (can? :read, sub) -style in whole file i think

jamox:
Initial thoughts; maybe this should be exercise level setting, so that the tmc-exam-exercise could be part of the course

app/controllers/courses_controller.rb
```diff
@@ -198,6 +198,13 @@ def manage_exercises
     @exercises_id_map = @exercises.map { |e| [e.id, e] }.to_h
   end
 
+  def toggle_submission_result_visibility
+    authorize! :toggle_submission_result_visibility, @course
+    @course.hide_submission_result = !@course.hide_submission_result
+    @course.save!
```
jamox: Nit: this could be a method in model...
```ruby
def toggle_submission_result_visiblity
  self.hide_submission_results = !self.hide_submission_results
end
```

app/controllers/submissions_controller.rb
```diff
@@ -3,7 +3,7 @@
 # Receives submissions and presents the full submission list and submission view.
 # Also handles rerun requests.
 class SubmissionsController < ApplicationController
-  around_action :course_transaction
+  around_action :course_transaction # is this really needed to do all methods. should have except: [:index, :show] ?
```
jamox: is the grammar/spelling correct ;)

app/models/submission.rb
```diff
@@ -110,6 +112,12 @@ def status
     end
   end
 
+  def cannot_see_results?(user)
+    (all_tests_passed? || tests_ran?) &&
```
jsopakar: Was just thinking, if this shouldn't be in negative form. most other similar kind of methods are asking positive result like visible_to, so should this be can_see_results? just an idea, not sure if it's better.. if this is more clear on calling.

app/controllers/submissions_controller.rb
```diff
@@ -44,47 +48,47 @@ def show
...
+            }
+            when :ok then {
+              test_cases: @submission.test_case_records,
+              feedback_questions: @course.feedback_questions.order(:position).map(&:record_for_api),
+              feedback_answer_url: submission_feedback_answers_url(@submission, format: :json),
+            }
+            when :fail then {
+              test_cases: @submission.test_case_records
+            }
+            when :hidden then {
```
jamox: this hidden feels like it the hidden field for course. I assume it really refers to the hide submission results. consider renaming it; now it could easily confuse endusers and developers

app/helpers/submissions_helper.rb
```diff
@@ -16,8 +16,9 @@ def show_submission_list(submissions, options = {})
   end
 
   def submission_status(submission)
-    status = submission.status
-    raw("<span class=\"#{status}\">#{status.to_s.capitalize}</span>")
+    status = submission.status(current_user)
+    # if status is :hidden we have to give something else to css so that we can see it
+    raw("<span class=\"#{status == :hidden ? 'h_i_d_d_e_n' : status}\">#{status.to_s.capitalize}</span>")
```
jamox: I don't really like the h_i_d_d_e_n...

app/models/ability.rb
```diff
@@ -60,6 +60,14 @@ def initialize(user)
         can? :teach, sub.course
       end
 
+      can :download, Submission do |sub|
+        !sub.course.hide_submission_result? && can?(:read, sub)
```
jamox: Why the downloading of a zip of the exercise depends on whether the course allows the user to view results?
(I assume it is the download for the zip)

app/models/submission.rb
```diff
@@ -98,9 +98,11 @@ def valid_param?(k, v)
     k =~ /^[a-zA-Z\-_]+$/ && v =~ /^[a-zA-Z\-_]+$/
   end
 
-  def status
+  def status(user)
     if !processed?
       :processing
+    elsif cannot_see_results?(user)
```
jamox: should it be ! can_see_results?

app/views/submissions/_details.html.erb
```diff
@@ -35,7 +35,7 @@
             <tr>
               <th>Case</th>
               <th>Status</th>
-              <% if !@submission.all_tests_passed? %>
+              <% if !@submission.all_tests_passed? && can?(:read_result, @submission) %>
```
jamox: Try to be consistent with the use of brackets with `can?`. The bracksts were just removed from app/views/participants/show.html.erb

(I'd prefer it w/o the brackets)

Termanty:
@jamox Teacher will turn the submission results of the course hidden when exam will start and returns back to normal when exam is over. I don't see any advantage of exercise specific visibility.

app/controllers/submissions_controller.rb
Termanty: This status is used many places in views to tell the status of the submission. Status :hidden works visually.

image

app/models/course.rb
```diff
@@ -250,6 +250,11 @@ def submissions_to_review
     submissions.where('(requests_review OR requires_review) AND NOT reviewed AND NOT newer_submission_reviewed AND NOT review_dismissed')
   end
 
+  def toggle_submission_result_visiblity
```
jamox: should it end with `!`?

jamox:
I still feel like this would make more sense if the hide_results would be per exercise; this would allow teacher to keep on using the main course for the machine exam - by hiding only the results of the machine exam exercises. This would remove the need for having a separate course for the machine exam.

jamox:
Redone in #324

PR #303, opened 2015-07-10
Organization with zero courses view adjustment & accept/reject link on organization page

jruo:


app/views/organizations/show.html.erb
```diff
@@ -1,5 +1,14 @@
 <h1><%= @organization.name %></h1>
 
+<% if @organization.acceptance_pending? %>
+    <p>
+      <strong>This organization needs to be accepted by an administrator.</strong>
+      <% if can? :view, :organization_requests %>
+          <%= link_to 'Show penging requests', list_requests_organizations_path %>
```
jamox: Hmm, why not just show accept/reject page here?

jsopakar: Yeah, true. Added straight buttons to do actual accept/reject, instead of the link only.

Should be ready for merge now I think.

jamox:
LGTM

jamox:
Merged manually to master

PR #304, opened 2015-07-10
Stats for organization

jruo:


app/views/stats/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<h1>Statistics</h1>
+<h1>Statistics<%=" for #{@organization.name}" unless @organization.nil?%> </h1>
```
jamox: Is this a hack? -- when we would have this page w/o org?

jamox:
LGTM if tests pass + comment above

app/views/stats/index.html.erb
jsopakar: Old functionality to show all courses (for all organizations) stats in /stats was still retained. This was also kept without permission checks, based on irc discussions.

Do we want to get rid of that? Even for admin purposes? Can change if needed.

mpartel: I think it's fine

jamox:
LGMT

jamox:
Cherry-picked yesterday and now in prod

PR #309, opened 2015-07-31
Submission handled to be late if submitted adter soft deadline

jsopakar:
Submissions, which are submitted late after soft deadline, but before hard deadline, are marked as late in awardedpoints. Also points page will show late points seperately when needed.

### DIFF
--- a/app/controllers/participants_controller.rb
+++ b/app/controllers/participants_controller.rb
@@ -49,7 +49,7 @@ class ParticipantsController < ApplicationController
   def show
     @user = User.find(params[:id])
     authorize! :read, @user
-    @awarded_points = Hash[@user.awarded_points.to_a.sort!.group_by(&:course_id).map { |k, v| [k, v.map(&:name)] }]
+    @awarded_points = @user.awarded_points.to_a.sort!.group_by(&:course_id)
 
     if current_user.administrator?
       add_breadcrumb @user.username, participant_path(@user)
--- a/app/helpers/points_helper.rb
+++ b/app/helpers/points_helper.rb
@@ -21,4 +21,27 @@ module PointsHelper
   def points_list(points)
     points.map { |pt| h(pt) }.join('&nbsp; ').html_safe
   end
+
+  def points_list_obj(points)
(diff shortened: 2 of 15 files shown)


app/controllers/participants_controller.rb
```diff
@@ -49,7 +49,7 @@ def index
   def show
     @user = User.find(params[:id])
     authorize! :read, @user
-    @awarded_points = Hash[@user.awarded_points.to_a.sort!.group_by(&:course_id).map { |k, v| [k, v.map(&:name)] }]
+    @awarded_points = @user.awarded_points.to_a.sort!.group_by(&:course_id)
```
jamox: Why did we sort them in an array, not in db...

app/views/points/index.html.erb
```diff
@@ -82,12 +90,48 @@
...
             <td>
-              <%= "#{user_points_for_this_sheet}/#{sheet[:total_available]}" %>
+              <% unless user_late_points_for_this_sheet == 0 %>
+                <div class="late-points-toggle" style="display: none;">
+                  <%=
+                      raw("#{user_points_for_this_sheet - user_late_points_for_this_sheet}" +
+                            '<span class="late-points">' +
+                            "+#{user_late_points_for_this_sheet}" +
+                            '</span>' +
+                            "/#{sheet[:total_available]}")
```
jamox: Let's make it not to use raw, but just build it as normal erb

app/helpers/points_helper.rb
```diff
@@ -21,4 +21,27 @@ def gdocs_notifications(notifications)
...
   end
+
+  def points_list_obj(points)
+    points.map do |pt|
+      if pt.late?
+        "<span class='late-points'>#{h(pt.name)}*</span>"
+      else
+        h(pt.name)
+      end
+    end.join('&nbsp; ').html_safe
```
jamox: check the html safe behaviour :)

jamox:
https://github.com/jamox/tmc-server/tree/late-points

PR #314, opened 2015-08-05
Better ui for managing exercises

jruo:


app/views/courses/manage_exercises.html.erb
```diff
@@ -1,4 +1,8 @@
-<h1>Enable and disable exercises</h1>
+<h1 class="float-left">Enable and disable exercises</h1>
+<div class="float-right" style="margin: 25px 20px 0 0">
```
jamox: inline styles :/

jamox:
LGTM, though the inline styles are not that nice

PR #318, opened 2015-08-10
URL with name instead of id

Termanty:
This pull request changes course id to course name in URL. Same thing is done also for exercise. 
Security hole is removed by introducing check for URL parameters so that teachers from one organization can't do changes to courses in  other organization.


jamox:
Just a thought - we might want to use this: http://guides.rubyonrails.org/routing.html#overriding-named-route-parameters so that we would have params[:name] instead of params[:id]

app/controllers/courses_controller.rb
```diff
@@ -206,10 +206,17 @@ def assign_show_view_vars
 
   def set_organization
     @organization = Organization.find_by(slug: params[:organization_id])
+    fail ActiveRecord::RecordNotFound unless @organization
```
jamox: find_by! maybe?

app/controllers/courses_controller.rb
```diff
@@ -206,10 +206,17 @@ def assign_show_view_vars
...
 
   def set_course
-    @course = Course.find(params[:id])
+    @course = Course.find_by(name: params[:id])
+    fail ActiveRecord::RecordNotFound unless @course
+    check_course_matches_organization
+  end
+
+  def check_course_matches_organization
+    fail ActiveRecord::RecordNotFound unless @course && @course.organization == @organization
```
jamox: nice!

app/controllers/emails_controller.rb
```diff
@@ -26,7 +26,7 @@ def index_global
   end
 
   def index_course
-    @course = Course.find(params[:id])
+    @course = Course.find_by(name: params[:id])
```
jamox: http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by-21

jruo:
Resolved conflicts and tests pass.

mpartel:
This doesn't preserve old URLs does it? This will almost certainly cause strange errors to users for a while after the update, until their clients refresh the course data. If we go ahead with this, the update should be done late at night IMO.

jamox:
Yes, I was thinking whether we really would like to keep supporting the use of id for these, and I didn't come up with any good reasons; thus I'd say only using name as the route param would be fine.

And without any changes on the server configuration this would most likely require all tmc-netbeans users to refresh the course list.
But I was thinking on generating apache url rewrite map to temporarily support old urls. Though I'm not yet sure whether it is worth generating that file...

mpartel:
Ah, good point. IMO requiring a course list refresh from each user is unacceptable. One option might be to patch the client to do this automatically if the course URL returns a 4xx.

I'd really like to avoid more "temporary" server-side hacks if possible.

jamox:
Yea, we definitely should not break the behavior with our server side changes. :)
That patch for the client would probably make sense, but first forcing an update out and then making this change seems to be of bigger cost, than the rewrite map this time. 

I'v suggested adding this behavior to tmc-core https://github.com/rage/tmc-core/issues/39

jamox:
I feel that it might just make more sense to keep on supporting our current routing schemes for a while; thus we should continue working on this to allow both kind of urls to be used for routing.

PR #327, opened 2015-08-12
Add the cookie consent banner as required by the EU

jamox:
WIP - RFC
eu

### DIFF
--- a/app/assets/stylesheets/main.css
+++ b/app/assets/stylesheets/main.css
@@ -377,3 +377,8 @@ tr>.align-right {
 .no-wrap {
   white-space: nowrap;
 }
+
+/* EU Cookie consent */
+.cookie_consent {
+  margin-bottom: 0px;
+}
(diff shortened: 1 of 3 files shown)


jamox:
@mpartel Please review -- I chose the http://cookiepedia.co.uk/ for the more info as it seems to be nicely treating the cookie consent thingy.

Also about the message - now it is pretty much the least amount of text as possible - should we maybe elaborate on this.

mpartel:
I think this is not necessary. http://ec.europa.eu/ipg/basics/legal/cookies/index_en.htm says e.g. "authentication cookies, to identify the user once he has logged in, for the duration of a session" are exempt from the requirement.

It will become necessary if we ever add analytics cookies or such.

mpartel:
I think the above is only sure* to apply when the login cookie is non-persistent, which ours is.

[*] this stuff is infuriatingly vague

jamox:
Closed as this turns out to be irrelevant. But kept safe should we ever add more tracking.

PR #330, opened 2015-08-13
Made repo validation to use git ls-remote

jsopakar:
Using git ls-remote to check repo validity instead of whole cloning. Timeout check is for situations where git command tries to ask login, and would be left waiting forever without.

### DIFF
--- a/app/models/course_template.rb
+++ b/app/models/course_template.rb
@@ -47,13 +47,24 @@ class CourseTemplate < ActiveRecord::Base
   end
 
   def valid_git_repo?
-    return true unless source_url_changed? || git_branch_changed? # don't attempt repo cloning if source url or branch wasn't even changed
-    Dir.mktmpdir do |dir|
-      sh!('git', 'clone', '-q', '-b', git_branch, source_url, dir)
-      File.exist?("#{dir}/.git")
+    safe_source_url = Shellwords.escape(source_url)
+    safe_git_branch = Shellwords.escape(git_branch)
+    cmd = "git ls-remote --exit-code #{safe_source_url} #{safe_git_branch}"
+    output = ''
+    begin
+      Timeout::timeout(5) do
+        output = `#{cmd} 2>&1`
+      end
+    rescue Timeout::Error
+      errors.add(:base, "Cannot clone repository, timeout expired, git ls-remote output: #{output}")
+      return false
     end
-  rescue StandardError => e
-    errors.add(:base, 'Cannot clone repository. Error: ' + e.to_s)
+    status = $?
+    unless status.success?
+      errors.add(:base, "Cannot clone repository, git ls-remote output: #{output}")
+      return false
+    end
+    true
   end
 
   def valid_source_backend?
--- a/app/models/course_template.rb
+++ b/app/models/course_template.rb
@@ -48,24 +48,10 @@ class CourseTemplate < ActiveRecord::Base
 
   def valid_git_repo?
     timeout = 10
-    safe_source_url = Shellwords.escape(source_url)
-    safe_git_branch = Shellwords.escape(git_branch)
-    cmd = "timeout #{timeout+1} git ls-remote --exit-code #{safe_source_url} #{safe_git_branch}"
-    output = ''
-    begin
-      Timeout::timeout(timeout) do
-        output = `#{cmd} 2>&1`
-      end
-    rescue Timeout::Error
-      errors.add(:base, "Cannot clone repository, timeout expired, git ls-remote output: #{output}")
-      return false
-    end
-    status = $?
-    unless status.success?
-      errors.add(:base, "Cannot clone repository, git ls-remote output: #{output}")
-      return false
-    end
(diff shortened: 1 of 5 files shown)


mpartel:
Does the timeout kill the command subprocess too?

jsopakar:
Good catch, it was left alive. Now it should be killed, after 1 more sec, to make sure to catch timeout always. While manually watching, the process was killed now properly.

app/models/course_template.rb
```diff
@@ -47,13 +47,25 @@ def self.new_dummy(course, options = {})
...
+    safe_git_branch = Shellwords.escape(git_branch)
+    cmd = "timeout #{timeout+1} git ls-remote --exit-code #{safe_source_url} #{safe_git_branch}"
+    output = ''
+    begin
+      Timeout::timeout(timeout) do
+        output = `#{cmd} 2>&1`
+      end
+    rescue Timeout::Error
+      errors.add(:base, "Cannot clone repository, timeout expired, git ls-remote output: #{output}")
+      return false
```
jamox: As discussed with @mpartel last week, we probably should move this to the SystemCommands class and keep all escapes there.

jsopakar:
Now the system command !sh includes option for timeout. Uses the same method as was earlier on this custom checks.

app/models/course_template.rb
```diff
@@ -47,11 +47,9 @@ def self.new_dummy(course, options = {})
   end
 
   def valid_git_repo?
-    return true unless source_url_changed? || git_branch_changed? # don't attempt repo cloning if source url or branch wasn't even changed
-    Dir.mktmpdir do |dir|
-      sh!('git', 'clone', '-q', '-b', git_branch, source_url, dir)
-      File.exist?("#{dir}/.git")
-    end
+    timeout = 10
```
mpartel: Unnecessary variable

app/models/course_template.rb
```diff
@@ -47,11 +47,9 @@ def self.new_dummy(course, options = {})
...
+    sh!('git', 'ls-remote', '--exit-code', source_url, git_branch, { timeout: timeout })
```
mpartel: No need for { and }

app/models/course_template.rb
```diff
@@ -47,11 +47,9 @@ def self.new_dummy(course, options = {})
...
+    true
   rescue StandardError => e
     errors.add(:base, 'Cannot clone repository. Error: ' + e.to_s)
```
mpartel: Also return false explicitly?

lib/system_commands.rb
```diff
@@ -16,10 +16,11 @@ def system!(cmd)
   def sh!(*args)
     options = {
       assert_silent: false,
-      escape: true
+      escape: true,
+      timeout: 0
```
mpartel: I'd prefer nil rather than 0 as the default.

lib/system_commands.rb
```diff
@@ -32,8 +33,24 @@ def sh!(*args)
...
 
-    output = `#{cmd} 2>&1`
-    status = $?
+    output = nil
+    status = nil
+
+    if options[:timeout] > 0
+      cmd = "timeout #{options[:timeout]+1} #{cmd}"
+      begin
+        Timeout::timeout(options[:timeout]) do
```
mpartel: Can we get rid of this (and the +1 thing above?) since the `timeout` command should fail (with exit status 124) when time runs out.

jsopakar:
Changes for previous requests made. Much simplier now with just status code checking. Tests passed.

lib/system_commands.rb
```diff
@@ -32,8 +33,14 @@ def sh!(*args)
...
     end
 
+    unless options[:timeout].nil?
+      cmd = "timeout #{options[:timeout]} #{cmd}"
+    end
+
     output = `#{cmd} 2>&1`
     status = $?
+
+    fail "Time out for command ''#{cmd}'', status #{status.inspect}. The output follows:\n#{output}" if status.exitstatus == 124
```
mpartel: I'd like to nitpick the wording a bit. How about:

    "Command `#{cmd}` timed out, status ..."

spec/lib/system_commands_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'spec_helper'
+
+describe SystemCommands do
+
+  it 'should return error on timeout properly' do
+    expect {
+      sh!('sleep', 2, { timeout: 1 })
+    }.to raise_error
```
mpartel: Could we assert that the exception text contains something about timing out?

spec/lib/system_commands_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  it 'should not return error without timeout set' do
```
mpartel: This seems unnececssary. I'd rather have a test that checks that an exception is raised if the command fails with a non-zero exit status.

mpartel:
LGTM after the above nitpicks.

jsopakar:
Those requests changed too. Didn't have time to run whole test suite for this now though after latest changes.

mpartel:
Thanks!

PR #332, opened 2015-08-14
Added organization support for participant list

jsopakar:
Participant index now supports organization too, list only student/courses for the organization, if it's given in path.

So /org/hy/participants/ should work now. This was needed from some external scoring system, asked by @josalmi

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -35,6 +35,13 @@ class User < ActiveRecord::Base
       .group('users.id')
   end
 
+  # TODO: Later after enrollment has implemented, this should use it instead
+  def self.organization_students(organization)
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -35,6 +35,13 @@ def self.course_students(course)
       .group('users.id')
   end
 
+  # TODO: Later after enrollment has implemented, this should use it instead
+  def self.organization_students(organization)
+    joins(awarded_points: :course)
+      .where(courses: { organization_id: organization.id })
+      .group('users.id')
```
jamox: can the group be done like where... (no quotes...)

PR #333, opened 2015-08-17
Email verification

Termanty:
Email verification for new and old users.

### DIFF
--- a/app/views/email_confirmation_mailer/confirmation_link_email.text.erb
+++ b/app/views/email_confirmation_mailer/confirmation_link_email.text.erb
@@ -1,0 +1,5 @@
+Hi <%= @user.username %>,
+
+To confirm your email address click the URL below. After confirmation you can sign in.
+
+<a href="<%= @url %>"><%= @url %></a>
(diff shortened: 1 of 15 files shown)


app/views/email_confirmation_mailer/confirmation_link_email.text.erb
```diff
@@ -0,0 +1,5 @@
+Hi <%= @user.username %>,
+
+To confirm your email address click the URL below. After confirmation you can sign in.
+
+<a href="<%= @url %>"><%= @url %></a>
```
jamox: It probably should be just plain text here

db/migrate/20150812094445_add_email_confirmed_at_to_user.rb
```diff
@@ -0,0 +1,6 @@
+class AddEmailConfirmedAtToUser < ActiveRecord::Migration
+  def change
+    add_column :users, :email_confirmed_at, :date
+    add_column :users, :confirm_token, :string
```
jamox: Maybe we should do it more like with the password `password_reset_keys` (or use the same table)

PR #334, opened 2015-08-18
Action token

Termanty:
This PR changes PasswordResetKey to more generic ActionToken, which can be used in email confirmation and possibly with course enrollments.


jamox:
LGTM, nice work!

PR #335, opened 2015-08-18
Move course options to ui

jruo:


lib/course_refresher.rb
```diff
@@ -189,6 +189,8 @@ def check_directory_names
     end
 
     def update_course_options
+      return if @course.refreshed?
```
jamox: Maybe comment why this 
`#after initial refresh all options are set and updated via web ui`

jamox:
LGTM, nice work

db/migrate/20150814085023_remove_old_course_hidden_fields.rb
```diff
@@ -0,0 +1,33 @@
+class RemoveOldCourseHiddenFields < ActiveRecord::Migration
+  def up
+    rename_column :courses, :disabled_status, :status
+
+    Course.all.each do |course|
```
jamox: Create the
```ruby
class Course < ActiveRecord::Base
end
```
to work around any possible changes to course

jamox:
This is waiting just for some manual testing from me, and I can easily fix the small comment.

PR #337, opened 2015-08-21
User email validation

jsopakar:


app/controllers/users_controller.rb
```diff
@@ -30,10 +30,11 @@ def create
     set_user_fields
 
     if @user.errors.empty? && @user.save
+      send_email_confirmation_mail(@user)
       if @bare_layout
-        render text: '<div class="success" style="font-size: 14pt; margin: 10pt;">User account created.</div>', layout: true
+        render text: '<div class="success" style="font-size: 14pt; margin: 10pt;">User account created. Please confirm your email address to continue.</div>', layout: true
       else
-        flash[:notice] = 'User account created. You can now log in.'
+        flash[:notice] = 'User account created. Please confirm your email address to continue.'
```
jsopakar: This could be bit more clear for user, clearly asking to check your emails. Bit like it was in another notice, which was like "Check your emails and click the confirmation link..."

app/models/action_token.rb
```diff
@@ -7,7 +7,7 @@
 # A password reset key is valid for a limited time and may safely be left unused.
```
jsopakar: Not actually for this PR changes, but initial comments for action tokens could be more general, those are propably left from old model.

app/controllers/users_controller.rb
```diff
@@ -30,10 +30,11 @@ def create
     set_user_fields
 
     if @user.errors.empty? && @user.save
+      send_email_confirmation_mail(@user)
       if @bare_layout
-        render text: '<div class="success" style="font-size: 14pt; margin: 10pt;">User account created.</div>', layout: true
+        render text: '<div class="success" style="font-size: 14pt; margin: 10pt;">User account created. Confirmation email has been sent to your email address.</div>', layout: true
       else
-        flash[:notice] = 'User account created. You can now log in.'
+        flash[:notice] = 'User account created. Confirmation email has been sent to your email address.'
```
jamox: We probably should throw in the mention to check the spam folder should they have problems receiving the email

app/controllers/sessions_controller.rb
```diff
@@ -14,6 +14,11 @@ def create
     redirect_params = {}
     if user.nil?
       redirect_params = { alert: 'Login or password incorrect. Try again.' }
+    elsif user.email_confirmed_at.nil?
+      non_confirmed user
+      @user = user
+      redirect_to email_confirmation_request_path
+      return
```
jamox: Should we have the return on the same line as the redirect
`return redirect_to email_confirmation_request_path`

app/controllers/sessions_controller.rb
```diff
@@ -14,6 +14,11 @@ def create
     redirect_params = {}
     if user.nil?
       redirect_params = { alert: 'Login or password incorrect. Try again.' }
+    elsif user.email_confirmed_at.nil?
```
jamox: So once in production we force automatically email validation?
Should we have a site setting for how to treat non new accounts?

app/models/action_token.rb
jamox: Yea, leftovers from the original rename. These were initially accurate, but we should make it more general...

app/models/user.rb
```diff
@@ -186,6 +186,21 @@ def assistantship(course)
     Assistantship.find_by(user_id: self, course_id: course)
   end
 
+  def activate_email
+    self.email_confirmed_at = Time.now
+    self.email_confirmation_token.destroy!
+    save!(:validate => false)
```
jamox: validate: false

though do we really need it... (skip validation)

jamox: should we prefer `self.update_attributes(email_confirmed_at: Time.now)`

app/models/user.rb
```diff
@@ -186,6 +186,21 @@ def assistantship(course)
...
+  end
+
+  def email_confirmation_token
+    self.action_tokens.find { |t| t.action == 'confirm_email' }
```
jamox: could we do this in db `self.action_tokens.where(action: 'confirm_email')`

app/views/email_confirmation_mailer/confirmation_link_email.text.erb
```diff
@@ -0,0 +1,5 @@
+Hi <%= @user.username %>,
+
+To confirm your email address click the URL below. After confirmation you can sign in.
+
+<a href="<%= @url %>"><%= @url %></a>
```
jamox: the `a` would look ugly in the plain text version

PR #338, opened 2015-08-21
Help texts for teacher ui, WIP

jsopakar:
Not meant to be ready yet, made for comments, and some todo post.

Started to add some more informative help texts for teacher ui. Css and layout things are just initially there.. could be tuned much better, suggestions are welcome.

Addition to these static explanations, some question mark working as tooltip would be nice in some situations. Didn't have yet time to check is there some built-in feature for this in bootstrap or something...

Also could link to full manual somewhere. Jamo had an idea to set the url in sitesettings.

### DIFF
--- a/app/controllers/course_templates_controller.rb
+++ b/app/controllers/course_templates_controller.rb
@@ -60,7 +60,7 @@ class CourseTemplatesController < ApplicationController
     @organization = Organization.find_by(slug: params[:organization_id])
     authorize! :teach, @organization
     add_organization_breadcrumb
-    add_breadcrumb 'Course templates'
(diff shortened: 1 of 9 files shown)


app/controllers/course_templates_controller.rb
```diff
@@ -60,7 +60,7 @@ def list_for_teachers
     @organization = Organization.find_by(slug: params[:organization_id])
     authorize! :teach, @organization
     add_organization_breadcrumb
-    add_breadcrumb 'Course templates'
+    add_breadcrumb 'Create new course'
```
jamox: For some reason I'd say "Create a new course"...

PR #342, opened 2015-09-04
Use another sandbox for tests, to allow running dev and test simultaneously

jamox:
Allows running tests and dev env simultaneously, by allowing tests to use any already running sandbox.

```shell
rake spec SANDBOX_HOST=127.0.0.1 SANDBOX_PORT=3001
```

This sure could cause tests/dev env to slow down and should be used with caution.
However, I feel this should improve the life in general.
All tests pass; tried running test suite once with the command above, when sandbox was run by the dev-env script

### DIFF
--- a/spec/support/remote_sandbox_for_testing.rb
+++ b/spec/support/remote_sandbox_for_testing.rb
@@ -5,6 +5,7 @@ class RemoteSandboxForTesting
   @server_pids = nil
   @result_queue = nil
   @server_ports = nil
+  @host = 'http://localhost'
 
   def self.server_ports
     # Currently just one server. Running multiple server webapps on the same machine is problematic.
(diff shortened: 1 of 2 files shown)


spec/support/remote_sandbox_for_testing.rb
```diff
@@ -36,12 +43,12 @@ def self.init_servers_as_root!(actual_user, actual_group)
...
   def self.init_stubs!
     RemoteSandbox.stub(:all) do
-      server_ports.map { |port| RemoteSandbox.new("http://localhost:#{port}/") }
+      server_ports.map { |port| RemoteSandbox.new("#{@host}:#{port}/") }
     end
   end
 
   def self.cleanup_after_all_tests!
-    @root_helper.stop
+    @root_helper.stop if @root_helper
```
jamox: This is kinda hacky... Not sure if I like this..

jamox:
Implemented, discarding this PR

PR #344, opened 2015-09-08
Travis run all tests on travis with a mock sandbox

jamox:
This is not absolutely the best way to get all tests to be run on travis, and it adds an extra step should the results/exercises change.

This also depends on my rather quick implementation of a [caching mock-sandbox](https://github.com/jamox/tmc-mock-sandbox).
w/o a PR reviewing/modifying it in github is kinda tricky; the Dir.chdir is not used as we can receive multiple parallel requests and I kept on getting warnings on that.

Let me know what you think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,19 @@
-sudo: true
+sudo: false
 
 language: ruby
+
 rvm:
   - 2.2.0
+
+addons:
+  apt:
+    sources:
+      - deadsnakes
+    packages:
+      - valgrind
+      - check
+      - pkg-config
+      - gcc
 
 git:
   submodules: false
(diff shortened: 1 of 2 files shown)


mpartel:
Sorry, I still really don't like the extra step and the added complexity and fragility. The gain just doesn't seem worth it to me :(

What are the remaining issues with running the real sandbox in Travis?

If building it is problematic, maybe we can host a prebuilt sandbox keyed on a git commit ID somewhere for Travis to download. If Travis cannot run UML at all then, well, I guess something like this is the least bad option, if we really want Travis.

Alternatively, we have servers right? Why not host our own more flexible CI server? I'm sure most of them have GitHub integration by now?

jamox:
I do actually mostly agree since this adds more fragility for the CI (which should be robust should we want to trust it).
Though it could make life easier for smaller fixes... though any other CI solution would do the trick...

One interesting thing I discovered while fiddling around with this idea is that we after all don't spend that much time in sandbox iff disk io is fast/normal. (speed improvement of the tests are minor)

I'll test whether we could with reasonable effort use a pre built sandbox on travis

mpartel:
Before putting much more effort into this, it might be worth profiling what the real bottlenecks are. It's surprising to me if the sandbox is not the largest one.

The only other obviously big one I can think of is the polling delay in the submission page waiting for the submission to get processed. Push notifications (with less frequent polling remaining as a fallback) would be the obvious fix there.

PR #354, opened 2015-10-26
Run all tests in Travis

jamox:
Run all* tests in travis with real sandbox.

We cannot use the container based setup, not beause we need sudo, which we don't, but because sandbox caused error `/dev/shm/ must be not mounted noexec`.

We also split running tests under three builds to gain reasonable test run speed, and it lets us to avoid the time limit of 50min for each task.

The build script does caching for bundler manually, maybe too manually but I'd leave it as it is for now.

* one test (the only maven on sandbox) is not being run, since to use maven it would require network access to download dependencies and use of tap devices is not possible in travis...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 sudo: true
 
+env:
+  - RSPEC_PATTERN="spec/{controllers,features,helpers,lib,mailers,models,usermanual}/**/*.rb"
+  - RSPEC_PATTERN="spec/integration/{running_tests,comet,feedback,requests}/**/*.rb"
+  - RSPEC_PATTERN="spec/integration/*.rb"
+
 language: ruby
 rvm:
   - 2.2.0
@@ -7,19 +12,34 @@ rvm:
 git:
   submodules: false
 
+addons:
+  apt:
+    packages:
+      - valgrind
+      - check
+      - pkg-config
+      - gcc
+
 before_install:
-  - sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
-  - git submodule update --init --recursive
-  - git config --global user.email "travis@example.com"
-  - git config --global user.name "Travis"
+  - . ./setup-travis.sh
+
+install:
+  - bundle install --retry=3 --jobs=3 --deployment
 
 services:
   - postgresql
 
 before_script:
   - createuser -U postgres -s tmc
+  - bundle exec rake db:reset
 
 script:
-  - bundle exec rake db:reset
-  - export rvmsudo_secure_path=1
-  - rvmsudo bundle exec rake spec SPEC="spec/controllers spec/models spec/helpers spec/mailers"
+  - ./ext/tmc-sandbox/web/webapp.rb start
+  - bundle exec rake spec SPEC_OPTS="--pattern $RSPEC_PATTERN --tag ~network -f d" SANDBOX_HOST=127.0.0.1 SANDBOX_PORT=3001
+
+after_failure:
+  - cat ext/tmc-sandbox/web/log/*
+
+notifications:
+  slack:
+    secure: YCiWybZYBoJ2JDjpPp5Idf4OvqScay5WfEuQMdvee42kgsSVolew6V5YUx41E29Of+2xAcQ+be5XImzO7SlTYuaGD6hf0JZvjjxO3gpycCljNSvDgFnBF9s4WuUQ31LozI6eQqo/9c5zUFxW9MGeeaRUWvurlW0rRHM3UKokAQk=
--- a/lib/maven_project.rb
+++ b/lib/maven_project.rb
@@ -58,7 +58,7 @@ class MavenProject
 
   def compile!
     Dir.chdir(path) do
-      SystemCommands.sh!('mvn -q package -Dmaven.test.skip=true', {escape: false})
+      SystemCommands.sh!('mvn', '-q', 'package', '-Dmaven.test.skip=true')
     end
   end
 
--- a/lib/maven_project.rb
+++ b/lib/maven_project.rb
@@ -58,7 +58,7 @@ class MavenProject
 
   def compile!
     Dir.chdir(path) do
-      SystemCommands.sh!('mvn', '-q', 'package')
+      SystemCommands.sh!('mvn -q package -Dmaven.test.skip=true', {escape: false})
     end
   end
 
--- a/lib/tasks/travis.rake
+++ b/lib/tasks/travis.rake
@@ -1,0 +1,15 @@
+require 'system_commands'
+
+namespace :travis do
+  desc "Update vendor/bundle files to gem cache on tmc-server"
+  task :gemcache do
+    unless Dir.exists? 'vendor/bundle'
+      puts 'No vendor/bundle exists, create it with `bundle install --path vendor/bundle`'
+      exit 1
+    end
+    puts "Pushing new gems from local gem cache (vendor/bundle) to travis cache"
+    SystemCommands.sh!("find vendor/bundle -iname '*.gem'  -print0  | rsync -0 --checksum -avz  --files-from=- . web01.testmycode.net:/srv/www/testmycode.net/travis/rubygems-cache/", {escape: false})
+    puts "Done"
+  end
+end
+
--- a/setup-travis.sh
+++ b/setup-travis.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+# Sets up dependencies and environment for tests to be run in Travis.
+#
+# This is not recommended to be run locally, and it employs some hacks and tricks to get by limitations of travis.
+
+set -ev
+
+# Unset bundle Gemfile, since we have multiple projects with separate Gemfiles and it will just cause confusion
+unset BUNDLE_GEMFILE
+
+# Update bundler
+gem update --system
+gem --version
+
+# Cache file containing vendor/bundle/cache - to speed up bundle install and to fix possible connection timeouts to Rubygems
+wget -r -q -nH --cut-dirs=2 --no-parent -A gem  http://testmycode.net/travis/rubygems-cache/vendor/
+bundle install --retry=3 --jobs=3 --deployment
+
+# Git settings so that Travis can clone submodules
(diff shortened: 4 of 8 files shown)


jamox:
Don't mind the errors shown in https://travis-ci.org/testmycode/tmc-server/jobs/87557545 -- I'll fix them once travis is up and running :)

jamox:
@mpartel a quick review would be nice

.travis.yml
```diff
@@ -7,19 +12,39 @@ rvm:
...
-  - rvmsudo bundle exec rake spec SPEC="spec/controllers spec/models spec/helpers spec/mailers"
+
+script:
+  - env
+  - BUNDLE_GEMFILE=$(pwd)/ext/tmc-sandbox/web/Gemfile ./ext/tmc-sandbox/web/webapp.rb start
+  - BUNDLE_GEMFILE=$(pwd)/ext/tmc-sandbox/web/Gemfile ./ext/tmc-sandbox/web/webapp.rb status
+  - sleep 10
+  - BUNDLE_GEMFILE=$(pwd)/ext/tmc-sandbox/web/Gemfile ./ext/tmc-sandbox/web/webapp.rb status
+  - curl 127.0.0.1:3001
+  - curl -v 127.0.0.1:3001/status.json
```
mpartel: Instead of the sleep, could we have e.g. a for-loop that checks `$(curl http://127.0.0.1:3001/status.json)` against an expected JSON string a limited number of times and waits for 1 sec between checks?

lib/maven_project.rb
```diff
@@ -58,7 +58,7 @@ def compiled?
 
   def compile!
     Dir.chdir(path) do
-      SystemCommands.sh!('mvn', '-q', 'package')
+      SystemCommands.sh!('mvn -q package -Dmaven.test.skip=true', {escape: false})
```
mpartel: nit: I wouldn't use `escape: false` without a good reason. Same below.

setup-travis.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+set -ev
+
+gem update --system
+gem --version
+#cache file containing vendor/bundle/cache - to speed up bundle instann and to fix possible connection itmeouts to rubygems
```
mpartel: please fix typos

setup-travis.sh
```diff
@@ -0,0 +1,45 @@
...
+wget -qO- http://testmycode.net/travis/2.2.0-tmc-server-bundler-cache.tar.gz | tar xz
+BUNDLE_GEMFILE=Gemfile bundle install --retry=3 --jobs=3 --deployment
```
mpartel: Maybe comment why setting BUNDLE_GEMFILE explicitly (here and below) is necessary?

setup-travis.sh
```diff
@@ -0,0 +1,45 @@
...
+# Reduce mavens memory usage
+export MAVEN_OPTS="-Xms512m -Xmx1024m -XX:PermSize=1024m"
+# So that travis can clone submodules
```
mpartel: At first this sounded like it's a continuation of the previous comment. Maybe say "Git settings so that ..."

setup-travis.sh
```diff
@@ -0,0 +1,45 @@
...
+sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
+git submodule update --init --recursive
+git config --global user.email "travis@example.com"
+git config --global user.name "Travis"
+  # tmc-check
```
mpartel: Be consistent about having a space before the comment.

setup-travis.sh
```diff
@@ -0,0 +1,45 @@
...
+cd tmc-check
+env BUNDLE_GEMFILE=$(pwd)/Gemfile bundle install --jobs=3 --retry=3 --deployment
+env BUNDLE_GEMFILE=$(pwd)/Gemfile make
+env BUNDLE_GEMFILE=$(pwd)/Gemfile make install PREFIX=$CHECK_INSTALL_DIR BUNDLE_GEMFILE="$(pwd)/Gemfile"
+export PATH="$CHECK_INSTALL_DIR/bin:$PATH"
+cd ..
+export LD_LIBRARY_PATH=$CHECK_INSTALL_DIR/lib:/lib:/usr/lib:/usr/local/lib:$LD_LIBRARY_PATH
+export C_INCLUDE_PATH=$CHECK_INSTALL_DIR/include:$C_INCLUDE_PATH
+export PKG_CONFIG_PATH=$CHECK_INSTALL_DIR/lib/pkgconfig:$PKG_CONFIG_PATH
+  #tmc comet, compile manually since rake task keeps on failing in travis...
```
mpartel: O_o strange. Maybe link to a paste/gist about how it fails?

setup-travis.sh
```diff
@@ -0,0 +1,45 @@
...
+cd ext/tmc-comet/
+mvn install -q -Dmaven.test.skip=true
+cd ../../
+  # Build submodule deps
+BUNDLE_GEMFILE=Gemfile bundle exec rake compile
+  # use prebuilt tmc-sandbox
+wget -qO- http://testmycode.net/travis/sandbox-$(git submodule status ext/tmc-sandbox | grep -E -o  "[0-9a-f]{40}").tar.gz | tar xvz -C ext/
+cd ext/tmc-sandbox/web
+sed -i "s/ tmc/ $(whoami)/" site.defaults.yml
```
mpartel: I think it'd be better to be explicit about the option name(s) here like you are on the next line.

setup-travis.sh
```diff
@@ -0,0 +1,45 @@
...
+sed -i 's/max_instances: [0-9]/max_instances: 2/' site.defaults.yml
```
mpartel: `[0-9]*`? (or with [-r and +](http://stackoverflow.com/questions/22099623/sed-plus-sign-doesnt-work))

setup-travis.sh
```diff
@@ -0,0 +1,45 @@
...
+sed -i "s/\(enabled: \)true/\1false/" site.defaults.yml
```
mpartel: Be consistent about repeating the option name vs using `\1`

setup-travis.sh
```diff
@@ -0,0 +1,45 @@
...
+sed -i '/^network:$/{$!{N;s/^\(network:\)\n\(  enabled: \)true$/\1\n\2false/;ty;P;D;:y}}' site.defaults.yml
```
mpartel: Woah..

Though I see what it does, it could maybe use a comment.

mpartel:
Nice job, must have been frustrating :)

Could we also have a rake task that lets people with the appropriate ssh key easily package and upload new versions of stuff to http://testmycode.net/travis/ ?

Otherwise LGTM aside from the nitpicks, assuming it works.

lib/maven_project.rb
jamox: the escape false is needed for the `-Dmaven.test.skip=true` -- it would escape the `=` -> `\=`. I'll see if I could put it to env so that we don't need to explicitly add it for each project.

mpartel: Ok, strange that it affects the result (`echo \=` prints just `=`). Maybe just leave a comment about it then.

setup-travis.sh
jamox: Thavis exports that; I'll try unsetting it and see if that works

setup-travis.sh
jamox: Nice, yeah I didn't like how it sounded

jamox:
I'll run a few tests, sorry for the additional travis spam related to this PR

lib/tasks/travis.rake
```diff
@@ -0,0 +1,9 @@
+require 'system_commands'
+
+namespace :travis do
+  desc "Update vendor/bundle files to gem cache on tmc-server"
+  task :gemcache do
+    SystemCommands.sh!("find vendor/bundle -iname '*.gem'  -print0  | rsync -0 --checksum -avz  --files-from=- . web01:/srv/www/testmycode.net/travis/rubygems-cache/", {escape: false})
```
jamox: @mpartel not sure if you had thought something more general for this... This will at least simplify the rubygems part (though it is currently missing the install gems to vendor/bundle folder)

lib/maven_project.rb
jamox: Ok, so it turns out that I was wrong and you were right - the `escape: false` is not required.

lib/tasks/travis.rake
```diff
@@ -0,0 +1,11 @@
+require 'system_commands'
+
+namespace :travis do
+  desc "Update vendor/bundle files to gem cache on tmc-server"
+  task :gemcache do
+    puts "Pushing new gems from local gem cache (vendor/bundle) to travis cache"
+    SystemCommands.sh!("find vendor/bundle -iname '*.gem'  -print0  | rsync -0 --checksum -avz  --files-from=- . web01:/srv/www/testmycode.net/travis/rubygems-cache/", {escape: false})
```
mpartel: I wouldn't rely on having the "web01" alias. Use the full domain name?

If gems don't get put to `vendor/bundle` by default, add at least a `puts` of instrunctions to make that happen.

lib/tasks/travis.rake
```diff
@@ -0,0 +1,11 @@
...
+    puts "Done"
+  end
+end
```
mpartel: Another task for uploading a prebuilt sandbox? And maybe a toplevel `task :travis` that does both?

jamox: I had initially thought that it would be done more manually and not necessarily locally... But creating that task might be worth it.

lib/tasks/travis.rake
jamox: Good points both - I feel that having the server address set here is too specific to our setup...
Thou I don't feel like it should go to site.yml either. Maybe setting the host from env, but it diminish ease of this.

Or maybe just have it specifying the server (web01.testmycode.net) and our path for now.

jamox:
I'll put to pushing new files to travis cache to a todo list, for now it's priority is not high enough. Merging.

Thanks for the review!

PR #362, opened 2016-01-02
Start using tmc-langs for processing files

jamox:
To track efforts towards migrating to use tmc-langs for processing files.

- [x] Find exercises
- [x] Scan exercise
- [x] Get test case methods
- [x] Get points
- [x] Make solutions
- [x] Make stubs
- [x] Prepare submission
- [x] Run tests in sandbox, except for makefile

And adds env variable `$MAVEN_HOME` to runtime dependency.


jamox:
It's almost done, some more manual testing with real data and it should be ready for merge

PR #367, opened 2016-02-01
Include timestamp to points

jamox:


mpartel:
Does this additional join make it notably slower in production?

jamox:
Not really. The page has always been slow. However I think this #369 should improve it.

PR #375, opened 2016-03-23
Changed organization accept to verify

jsopakar:
Organization should be usable right away without waiting for admin accept. Model answers are hidden though.


app/controllers/solutions_controller.rb
```diff
@@ -15,6 +15,8 @@ def show
     rescue CanCan::AccessDenied
```
jamox: Sidenote: exceptions are slow. Let's refactor this in another pr...

spec/factories.rb
```diff
@@ -149,13 +149,13 @@ def make_repo_for_course_template
     sequence(:name) { |n| "organization#{n}" }
     sequence(:information) { |n| "information#{n}" }
     sequence(:slug) { |n| "organization#{n}" }
-    acceptance_pending true
+    verified false
   end
 
   factory :accepted_organization, class: Organization do
```
jamox: verified_organization and unverified_organization maybe

PR #382, opened 2016-04-19
Minor adjusts: teacher/assistant views unified, using course titles instead of names

jsopakar:


jamox:
LGTM, let's wait for the travis to pass.

PR #387, opened 2016-06-21
Setup wizard, first phase

jsopakar:
First step to easier, wizard type setup functionality. Main thing is to make course creation easier.

For now, some of the edit options and code are still in old place too, but goal is to move most of the setup things under the setup-namespace during next weeks.

Having problems with sandbox-related tests on development computer, so would be wise to run whole test suite once through.


jamox:
Awesome. Great work!

I had one test failing locally; I'll see if I can fix it.

jsopakar:
Forgot initially the usermanual, that test was failing. It's now updated to match the current wizard, and test should pass.

PR #397, opened 2016-08-16
Allow running tests in docker containers

jamox:
Uses for test result collection: https://github.com/jamox/remote_rspec_aggregator


jamox:
I'm going to merge this now and let's keep on optimizing this later on.

PR #400, opened 2016-08-26
Org and course pages teacher/admin functions restructured

jsopakar:
Especially course page was very messy with all kind of links and buttons. Here is a simple restructuring to have own panel for teacher and admin functions, for both organization and course pages. There's some rough css for panels, but colors etc can be tuned better later if needed.

This PR is built on top of #398, including those earlier changes. And some integration tests couldn't be run on my setup, sp please run all tests before mergeing.


jsopakar:
Example picture from course page:

image

jsopakar:
And for organization page:

image

PR #428, opened 2016-10-09
Prettier login screen

nygrenh:
Make the login screen a bit better so that we don't scare off our OAuth users.


nygrenh:
Before: 
tmc-login-before

After: 
tmc-login-after

PR #448, opened 2016-10-24
Add basecontroller to v8

Rochet2:
Adds the basecontroller to the v8.
This file is used by most (all?) other controllers in the v8.

@testmycode/api-team 

http://ohtu-jenkins.testmycode.io/job/v8basecontroller/4/

### DIFF
--- a/app/controllers/api/v8/base_controller.rb
+++ b/app/controllers/api/v8/base_controller.rb
@@ -7,8 +7,6 @@ class Api::V8::BaseController < ApplicationController
   #  before_action :doorkeeper_authorize!
   before_action :authenticate_user!
 
-  skip_authorization_check # we use doorkeeper in api, so let's skip cancancan
-
   rescue_from CanCan::AccessDenied do |e|
     render json: errors_json(e.message), status: :forbidden
   end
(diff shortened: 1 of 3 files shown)


app/controllers/api/v8/base_controller.rb
```diff
@@ -0,0 +1,44 @@
+require 'json'
+
+class Api::V7::BaseController < ApplicationController
+  clear_respond_to
+  respond_to :json
+
+  #  before_action :doorkeeper_authorize!
+  before_action :authenticate_user!
+
+  skip_authorization_check # we use doorkeeper in api, so let's skip cancancan
```
nygrenh: Do we want this? Cancancan can most likely work with doorkeeper tokens.

app/controllers/api/v8/base_controller.rb
```diff
@@ -0,0 +1,44 @@
+require 'json'
+
+class Api::V7::BaseController < ApplicationController
+  clear_respond_to
+  respond_to :json
+
+  #  before_action :doorkeeper_authorize!
+  before_action :authenticate_user!
```
nygrenh: Should we test this?

nygrenh: You can write controller tests and make assertions on the `@current_user` instance variable. I think it's important this always works correctly :).

Rochet2:
Jenkins succeeded http://ohtu-jenkins.testmycode.io/job/v8basecontroller/4/

PR #450, opened 2016-11-01
Api find exercises by course id or name

makevertti:
by id: /api/v8/courses/{course_id}/exercises
by name: /api/v8/organizations/{organization_id}/courses/{course_name}/exercises

Jenkins build: http://ohtu-jenkins.testmycode.io/job/tmc-server-api-exercise-search/4/
closes #431 
closes #432
@testmycode/api-team

PR #451, opened 2016-11-07
removed unnecessary query

makevertti:
Fix requested by @jamo 
https://github.com/testmycode/tmc-server/pull/450

jenkins build: http://ohtu-jenkins.testmycode.io/job/tmc-server-api-exercise-search/6/

### DIFF
--- a/app/controllers/api/v8/exercises_controller.rb
+++ b/app/controllers/api/v8/exercises_controller.rb
@@ -72,13 +72,11 @@ class Api::V8::ExercisesController < Api::V8::BaseController
       respond_not_found('Course not found!')
     else
       exercises = Exercise.where(course_id: course.id)
-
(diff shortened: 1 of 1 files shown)


app/controllers/api/v8/exercises_controller.rb
```diff
@@ -72,13 +72,11 @@ def get_by_course
...
     else
       exercises = Exercise.where(course_id: course.id)
-
       visible = exercises.select { |ex| ex.visible_to?(current_user) }
-
       presentable = visible.map do |ex|
         {
             id: ex.id,
-            available_points: Exercise.find_by(id: ex.id).available_points,
+            available_points: ex.available_points,
```
jamo: This helps, but not by much...
The `available_points` likely still causes a sql query, try with `includes` http://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations

PR #453, opened 2016-11-08
Api exercises download without authentication

### LINKED ISSUE #401: [API] Download exercises by name without authentication
An unauthenticated user should be able to download an exercise by knowing its name.

makevertti:
closes #401 
Jenkins build: http://ohtu-jenkins.testmycode.io/job/tmc-server-api-exercises-download/13/

### DIFF
--- a/app/controllers/api/v8/exercises_controller.rb
+++ b/app/controllers/api/v8/exercises_controller.rb
@@ -113,7 +113,6 @@ class Api::V8::ExercisesController < Api::V8::BaseController
   end
 
   def download
-    #self.class.skip_authorization_check
     course = Course.find_by!(name: "#{params[:slug]}-#{params[:name]}")
     exercise = Exercise.find_by!(name: params[:exercise_name], course_id: course.id)
 
--- a/app/controllers/api/v8/exercises_controller.rb
+++ b/app/controllers/api/v8/exercises_controller.rb
@@ -91,40 +91,33 @@ class Api::V8::ExercisesController < Api::V8::BaseController
 
   def get_by_course
     unauthorized_guest! if current_user.guest?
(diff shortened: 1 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -124,6 +124,11 @@ def respond_not_found(msg = 'Not Found')
     respond_with_error(msg, 404)
   end
 
+  def respond_not_found_with_auth_bypass(msg = 'Not Found')
+    authorize! :read, nil
```
nygrenh: `skip_authorization_check`

jackrl: Didn't we already agree that it doesn't work?

jackrl: We could maybe add a cancan definition for errors and use that instead

nygrenh: I remember that we concluded `skip_authorization_check` shouldn't be used  as a before action but I have no idea why it wouldn't work when called explicitly.

app/controllers/api/v8/exercises_controller.rb
```diff
@@ -64,28 +64,59 @@ class Api::V8::ExercisesController < Api::V8::BaseController
...
-            solution_visible_after: ex.solution_visible_after,
-            deadline: ex.deadline_for(current_user),
-            disabled: ex.disabled?
-        }
-      end
-      authorize! :read, visible
-      present(presentable)
+    course = Course.find_by!(id: params[:id]) if (params[:id])
+    course ||= Course.find_by!(name: "#{params[:slug]}-#{params[:name]}")
+    exercises = Exercise.where(course_id: course.id)
```
nygrenh: What happened to `.includes(:available_points)`?

jackrl: Fixed. Must have been removed on rebase by mistake.

PR #457, opened 2016-11-17
Split refresh

jsopakar:
Full refresh on course create is slow because available points calculation takes long time. Now refresh supports ignoring slow operations, and when new course is made, first refresh is made without that slow part. Instead, full refresh is launched on own thread to finish later at background. Course also has information if the first initial refresh is done or not, to disable submissions etc. until available points are calculated.

Splitting the refresh allows the new course wizard to progress fast and show course timings without minutes of waiting and even possible server timeout.

For admins, the front page shows if there are any courses which are not successfully initially refreshed, so potential problems can be found.

Another update in this PR is on course timings, where unlock options are offered better from drop down menu.

### DIFF
--- a/app/controllers/setup/course_details_controller.rb
+++ b/app/controllers/setup/course_details_controller.rb
@@ -37,18 +37,7 @@ class Setup::CourseDetailsController < Setup::SetupController
     if @course.save
       # Fast refresh without time-consuming tasks, like making solutions
       refresh_course(@course, no_directory_changes: @course.course_template.cache_exists?, no_background_operations: true)
-
-      # Full refresh on background
-      Thread.new do
-        begin
-          logger.info("Starting background initial refresh on course #{@course.name} (id #{@course.id})")
-          refresh_course(@course, no_directory_changes: @course.course_template.cache_exists?)
-          @course.initial_refresh_ready = true
-          @course.save!
-        rescue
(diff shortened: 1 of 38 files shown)


app/controllers/setup/course_details_controller.rb
```diff
@@ -35,7 +35,21 @@ def create
     @course.organization = @organization
 
     if @course.save
-      refresh_course(@course, no_directory_changes: @course.course_template.cache_exists?)
+      # Fast refresh without time-consuming tasks, like making solutions
+      refresh_course(@course, no_directory_changes: @course.course_template.cache_exists?, no_background_operations: true)
+
+      # Full refresh on background
+      Thread.new do
```
nygrenh: Background task

app/controllers/setup/course_timings_controller.rb
```diff
@@ -129,6 +140,14 @@ def fill_all_deadlines_with(date)
 
   private
 
+  def groups_as_array
```
nygrenh:
```
@course.exercise_groups.each_with_object(['']) do |group, array|
  array << group.name
end
```

jsopakar:
Changed background refresh to be done as background task, similar to submissions resendings are done. Also other comments are fixed.

Still running soon all tests to make sure everything works..

PR #468, opened 2017-03-17
Faster Course Pages

nygrenh:


mpartel:
Would it be cleaner and simpler to set `unlock_spec` to `nil` whenever we would now store an `unlock_spec` that results in `unlock_spec_obj.empty?`, and define `requires_unlock?` as `!unlock_spec.nil?`.

nygrenh:
@mpartel Good idea!

PR #471, opened 2017-08-09
Add created_at attribute to awarded point

juuha:
Created a new migration for adding created_at attribute to awarded point. We need it for calculating the average of points over time.

### DIFF
--- a/app/models/available_point.rb
+++ b/app/models/available_point.rb
@@ -85,7 +85,8 @@ class AvailablePoint < ActiveRecord::Base
       course_id: exercise.course_id,
       name: name,
       user_id: user.id,
-      submission: submission
+      submission: submission,
+      created_at: submission.created_at
     )
   rescue ActiveRecord::RecordNotUnique
   end
--- a/app/models/awarded_point.rb
+++ b/app/models/awarded_point.rb
@@ -8,13 +8,14 @@ class AwardedPoint < ActiveRecord::Base
   include Swagger::Blocks
 
   swagger_schema :AwardedPoint do
-    key :required, [:id, :course_id, :user_id, :submission_id, :name]
+    key :required, [:id, :course_id, :user_id, :submission_id, :name, :created_at]
 
     property :id, type: :integer, example: 1
     property :course_id, type: :integer, example: 1
     property :user_id, type: :integer, example: 1
     property :submission_id, type: :integer, example: 2
     property :name, type: :string, example: 'point name'
+    property :created_at, type: :string, example: '2016-10-17T11:10:17.295+03:00'
   end
 
   def point_as_json
@@ -23,7 +24,8 @@ class AwardedPoint < ActiveRecord::Base
       :course_id,
       :user_id,
       :submission_id,
-      :name
+      :name,
+      :created_at
     ]
   end
 
--- a/db/migrate/20170809122812_add_created_at_to_awarded_point.rb
+++ b/db/migrate/20170809122812_add_created_at_to_awarded_point.rb
@@ -1,8 +1,8 @@
 class AddCreatedAtToAwardedPoint < ActiveRecord::Migration
   def change
     add_column :awarded_points, :created_at, :datetime
-    AwardedPoint.all.each do |awarded_point|
-      awarded_point.update_attributes! :created_at => awarded_point.submission.created_at
+    AwardedPoint.find_each do |awarded_point|
+      awarded_point.update!(created_at: awarded_point.submission.created_at)
     end
   end
 end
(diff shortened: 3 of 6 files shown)


jamo:
Can't get just get the time from the submission

juuha:
A user can get their own submissions, but is denied when requesting other peoples' submissions. However, the user can request other peoples' points and receive them. We originally planned on requesting the timestamp from submissions, but ended up with this solution because of the privacy issue. 

To clarify, we are trying to calculate the average point gain over time of all users, not just from the currently logged in user.

jamo:
But can't we join Submissions with AwardedPoints and use that result for building buckets (possibly in sql level)

db/migrate/20170809125714_add_created_at_to_awarded_point.rb
```diff
@@ -0,0 +1,5 @@
+class AddCreatedAtToAwardedPoint < ActiveRecord::Migration
+  def change
+    add_column(:awarded_points, :created_at, :datetime)
```
nygrenh: It looks like you're not updating old records...

db/migrate/20170809125714_add_created_at_to_awarded_point.rb
```diff
@@ -0,0 +1,5 @@
+class AddCreatedAtToAwardedPoint < ActiveRecord::Migration
```
nygrenh: Is there a reason why the schema isn't updated as well? Did you run this migration?

juuha:
The change now includes the schema change and now affects existing records properly. It updates the awarded point records to include the created_at attribute and the value for it is received from the submission of the awarded point.

db/migrate/20170809122812_add_created_at_to_awarded_point.rb
```diff
@@ -0,0 +1,8 @@
+class AddCreatedAtToAwardedPoint < ActiveRecord::Migration
+  def change
+    add_column :awarded_points, :created_at, :datetime
+    AwardedPoint.all.each do |awarded_point|
+      awarded_point.update_attributes! :created_at => awarded_point.submission.created_at
```
nygrenh: `awarded_point.update!(created_at: awarded_point.submission.created_at)`

db/migrate/20170809122812_add_created_at_to_awarded_point.rb
```diff
@@ -0,0 +1,8 @@
+class AddCreatedAtToAwardedPoint < ActiveRecord::Migration
+  def change
+    add_column :awarded_points, :created_at, :datetime
+    AwardedPoint.find_each do |awarded_point|
+      awarded_point.update!(created_at: awarded_point.submission.created_at)
```
nygrenh: Actually submissions can be deleted and many for many awarded points submission is actually null. So this wont work on our production server :stuck_out_tongue:.

Maybe we should change this to:
```ruby
awarded_point.update!(created_at: awarded_point.submission.created_at) if awarded_point.submission
```

juuha:
Now the old awarded points with their submissions deleted, will remain with a nil value in the created_at column.

app/models/available_point.rb
```diff
@@ -85,7 +85,8 @@ def award_to(user, submission = nil)
       course_id: exercise.course_id,
       name: name,
       user_id: user.id,
-      submission: submission
+      submission: submission,
+      created_at: submission.created_at
```
Redande: Please add a comma after the last element

app/models/awarded_point.rb
```diff
@@ -23,7 +24,8 @@ def point_as_json
       :course_id,
       :user_id,
       :submission_id,
-      :name
+      :name,
+      :created_at
```
Redande: Comma here as well please

db/schema.rb
```diff
@@ -43,10 +43,11 @@
...
   create_table "awarded_points", force: :cascade do |t|
-    t.integer "course_id",     null: false
-    t.integer "user_id",       null: false
-    t.integer "submission_id"
-    t.string  "name",          null: false
+    t.integer  "course_id",     null: false
+    t.integer  "user_id",       null: false
+    t.integer  "submission_id"
+    t.string   "name",          null: false
+    t.datetime "created_at"
```
Redande: Also here a comma after the last element would be nice

app/models/awarded_point.rb
juuha: This one didn't have a comma either.

db/schema.rb
juuha: Here there was only a comma if there was further details about the element. Should I still place a comma?

app/models/available_point.rb
Redande: Actually, it would seem that ruby styleguide (https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas) doesn't like trailing commas, so no need to add them anywhere :stuck_out_tongue:

PR #473, opened 2017-11-30
Add endpoint for the last hour's submissions

nygrenh:


app/controllers/api/v8/courses/submissions/last_hour_controller.rb
```diff
@@ -0,0 +1,41 @@
+module Api
+  module V8
+    module Courses
+      module Submissions
+        class LastHourController < Api::V8::BaseController
```
Redande: This class should be added to ApidocsController so that the documentation is visible in documentation -endpoint

app/controllers/api/v8/courses/submissions/last_hour_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+                  end
+                end
+              end
+            end
+          end
+
+          def index
+            course = Course.find_by!(id: params[:course_id])
+            authorize! :read, course
+            return respond_access_denied unless current_user.administrator?
```
Redande: Should this be at the beginning of method since course-variable won't be used if the user isn't an admin?

app/controllers/api/v8/courses/submissions/last_hour_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+                  end
+                end
+              end
+            end
+          end
+
+          def index
+            course = Course.find(params[:course_id])
+            authorize! :read, course
+            return respond_access_denied unless current_user.administrator?
```
Redande: Should this admin check be at the beginning of the method since nothing is done with the course if the user isn't an admin?

app/controllers/api/v8/courses/submissions/last_hour_controller.rb
```diff
@@ -0,0 +1,41 @@
+module Api
+  module V8
+    module Courses
+      module Submissions
+        class LastHourController < Api::V8::BaseController
+          include Swagger::Blocks
+
+          swagger_path '/api/v8/courses/{course_id}/submissions' do
```
Redande: Should the path have "/last_hour" after submissions?

app/controllers/api/v8/apidocs_controller.rb
```diff
@@ -150,6 +150,7 @@ class ApidocsController < ActionController::Base
         Api::V8::Courses::ExercisesController,
         Api::V8::Courses::Exercises::PointsController,
         Api::V8::Courses::Exercises::Users::PointsController,
+        Api::V8::Courses::Exercises::Submissions::LastHourController,
```
Redande: This is incorrect, there's no Exercises module in LastHourController

PR #527, opened 2021-04-20
Handle internal sandbox errors with submissions

jolampi:
Set error originating from sandbox like `Docker image was not whitelisted.` as submission result to prevent infinite submission reprocessing.

~~This does not resolve the issue of containers failing to run in local environment due to the above error.~~

### DIFF
--- a/app/models/remote_sandbox.rb
+++ b/app/models/remote_sandbox.rb
@@ -11,6 +11,14 @@ require 'rust_langs_cli_executor'
 class RemoteSandbox
   attr_reader :baseurl
 
+  class InternalSandboxError < StandardError
+    attr_reader :object
+
+    def initialize(object)
+      @object = object
+    end
+  end
+
   class SandboxUnavailableError < StandardError; end
 
   def initialize(baseurl, experimental = true)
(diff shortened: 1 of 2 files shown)


app/models/remote_sandbox.rb
```diff
@@ -76,10 +84,21 @@ def send_submission(submission, notify_url)
         submission.save!
       rescue StandardError => e
         puts e
-        raise SandboxUnavailableError
+        begin
+          # Try to parse the response to see if the error originated from within the sandbox.
+          @parsed = JSON.parse e.response
+        rescue StandardError
+          raise SandboxUnavailableError
```
nygrenh, line 91: Make sure submission is tried again if the sandbox is full: https://github.com/rage/sandbox/blob/2e1c4d754cde345acd9bf9b3fd51ae578ea5195c/src/middleware/error_handler.ts#L35

app/models/remote_sandbox.rb
```diff
@@ -76,10 +84,21 @@ def send_submission(submission, notify_url)
...
+        end
+        raise InternalSandboxError.new(@parsed['error'])
       end
     rescue SandboxUnavailableError
       raise
+    rescue InternalSandboxError => e
+      submission.pretest_error = e.object
+      submission.processed = true
```
nygrenh, line 99: Cannot set processed true here because it will stop the retries.

Remove this and then in the retry  handler make sure that it does not override the pretest error set here.

https://github.com/testmycode/tmc-server/blob/master/lib/submission_processor.rb#L30

The  error message submission processor should be prepended to the pretest_error like
`Tried to process #{Submission.max_attempts_at_processing} times but failed. This is a system error.\n#{previous_pretest_error}`

nygrenh (changes requested):

nygrenh (approved):

PR #528, opened 2021-04-28
Update usermanual

Redande:


Redande:
Deployed to http://testmycode.github.io/tmc-server/usermanual/

PR #547, opened 2022-01-01
bugfix for firefox users

### LINKED ISSUE #548: tmc-paste: newlines inserted between each line of code copied from firefox
Copying from a tmc-paste code listing in Firefox inserts an extra newline between each line of code when pasted into a text editor.

Copying works as expected in the Chromium browser.

The problem disappears when highlightjs-line-numbers is prevented from loading. If line numbering is unimportant this may be the better solution.

The issue appears to be caused independently by to two issues with Firefox, at least one of which is a bug:
- Firefox inserts a line break before div block elements.
  - Can be fixed by setting CSS display: block-inline on the relevant divs
- Firefox inserts newlines when selecting elements with CSS property user-select: none
  - Seems to be related to https://bugzilla.mozilla.org/show_bug.cgi?id=1625534
  - Can be fixed by utilising a CSS hack to unset the user-select: none property for selector td.hljs-ln-numbers for Mozilla users only.

maikkeli:
BUG: Firefox inserts newlines between each line when code is copied from a code listing in a TMC paste page.
FIX: Two problems in the CSS, each of which independently cause the above issue.

Note: This fix has only been tested using local copies of the generated html pages -- I haven't spun up an instance of tmc-server to properly test it. This might not even be the correct place in the source tree to fix it.

Fixes #548

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -448,29 +448,9 @@ td.hljs-ln-numbers {
 	/* your custom style here */
 }
 
-/* 2022-01-01
-   BUG: Mozilla inserts newlines in text copied from code listing [https://bugzilla.mozilla.org/show_bug.cgi?id=1625534]
-   FIX: unset *user-select on mozilla browsers
-   (Uses CSS-hack described: [https://bugzilla.mozilla.org/show_bug.cgi?id=1446470]) */
-@-moz-document url-prefix() {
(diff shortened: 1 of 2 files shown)


Redande:
Hey, thank you for this pull request, this is a very good quality of life improvement!

I moved the styling to the correct location, otherwise this works, so I merged this PR.

PR #575, opened 2023-03-20
Fix incorrect pluralization in toast

ljleppan:
**NB: Not tested locally, please verify before merging.**

This changes the "new organization to verify" toast visible to admins from "17 new organization to verifies" to grammatically more correct "17 new organizations to verify".


Redande:
Tested locally, works like a charm. Thank you!

